PR #161, opened 2015-07-11
Tests passing in OS X 10.11b (15A215h)

impressiver:
Had to update a few GCC flags to get tests built/passing on OS X 10.11 public beta. Let me know if you'd prefer these flags be added conditionally based on GCC version and/or platform.

Thanks!


PetteriAimonen:
Hmm, is this with an actual GCC or with Clang? What is the GCC version?

impressiver:
```
$ gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 6.1.0 (clang-602.0.53) (based on LLVM 3.6.0svn)
Target: x86_64-apple-darwin15.0.0
Thread model: posix
```

I installed the OS X 10.11 beta, updated XCode/homebrew and whacked warnings until it compiled. So I'm very sure you have a better idea than myself on the correct updates :)

PetteriAimonen:
Yeah, so actually it is clang and not gcc. Currently the build rules just check the specified compiler name to figure out what options to give it, which fails a bit on Mac OS X where "gcc" is actually just an alias for clang.

Try with: `scons CC=clang CXX=clang`

Though the issue about "extended offsetof" might actually be worth fixing in nanopb, in case some other compiler is very strict on interpretation of C standard.

impressiver:
I did some Googling and hadn't realized Apple started aliasing clang as gcc since 10.9. Probably didn't need to know because I'd been installing gcc via brew since forever. So this is looking even more like the wrong way to skin the cat, but worth noting that OS X masquerades clang as gcc.

impressiver:
you beat me :)

PetteriAimonen:
Added note about this in README now.

PR #169, opened 2015-09-21
Add proper Python3 support to the generator

kylemanna:
This pull request attempts to port `nanopb_generator.py` to a python2 *and* python3 compatible version.

Arch Linux already defaults to python3 and Ubuntu and Debian are forthcoming.  My Arch Linux install had a python-protobuf build for python3, but not for python2.  As a result, none of the hacks to force python2 worked.

I've tested this against the unit tests and it seems good for both python2 and python3.

Test systems for python3: Arch Linux up to date as of today

* clang 3.6.2-4
* cmake 3.3.2-2
* protobuf 2.6.1-1
* python 3.4.3-3
* python-protobuf 3.0.0a3-2
* scons 2.3.6-1

Python 2 testing was in a docker container running Ubuntu:14.04:

* cmake  2.8.12.2-0ubuntu3
* gcc 4:4.8.2-1ubuntu6
* protobuf 2.6.1 (built from source as 2.5.x in apt has issues with oneof it appears)
* python 2.7.5-5ubuntu3
* python-protobuf 2.6.1 (built and installed via pip)
* scons 2.3.0-2

All of the ASCII string vs unicode stuff is a complete headache and I'm not convinced it's 100% correct in python2, but all unit tests at least pass.

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -175,7 +175,7 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
     add_custom_command(
       OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.c"
              "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb.h"
-      COMMAND ${PYTHON2_EXECUTABLE}
+      COMMAND ${PYTHON_EXECUTABLE}
       ARGS ${NANOPB_GENERATOR_EXECUTABLE} ${FIL_WE}.pb ${NANOPB_OPTIONS}
       DEPENDS "${CMAKE_CURRENT_BINARY_DIR}/${FIL_WE}.pb"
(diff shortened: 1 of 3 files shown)


PetteriAimonen:
Ah, protobuf now builds also against python3; yeah, time to add support to nanopb also.

I'll look into this after 0.3.4 is out (next week). I'll have to build a python3 based container to run the tests on.

kylemanna:
Probably the most important thing is to make sure that I didn't break python2 support on accident.  My testing suggests not.

PetteriAimonen:
Yeah, I trust the tests to catch most breakages (I'll of course review the changes also). But continuous python 3 testing machine will be needed to make sure that I won't break python 3 support in the future :)

kylemanna:
> But continuous python 3 testing machine will be needed to make sure that I won't break python 3 support in the future :)

I created a pull request for an automated build system that attempts to test both python2 and python3 and a number of compilers.  Check out #170

PR #171, opened 2015-09-22
decode: Fix compiler issue with gcc-5

PetteriAimonen:
* gcc 5.0 and 5.1 appear to take issue with this line and generates the
  following error:

    /home/nitro/tmp/nanopb/pb_decode.c: In function ‘pb_decode_noinit’:
    /home/nitro/tmp/nanopb/pb_decode.c:889:60: error: conversion to ‘uint8_t {aka unsigned char}’ from ‘int’ may alter its value [-Werror=conversion]
                 fields_seen[iter.required_field_index >> 3] |= (uint8_t)(1 << (iter.required_field_index & 7));
                                                                ^
* This seems like a compiler bug, but this workaround is harmless.


kylemanna:
Ha, diggin' out the good stuff? :)

PetteriAimonen:
@kylemanna Yeah, not brave enough to pick the more invasive changes right before 0.3.4 release, but this seemed safe enough :)

PR #176, opened 2015-10-08
generator: Remove trailing whitespace

kylemanna:
* Remove trailing whitespace
* No functional changes

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -86,7 +86,7 @@ class Names:
         if isinstance(parts, Names):
             parts = parts.parts
         self.parts = tuple(parts)
-    
+
     def __str__(self):
         return '_'.join(self.parts)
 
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Ah, trailing whitespace, what an annoying problem. I fear that I'll be adding it back pretty soon, as I tend to edit with random editors that are not always "properly" configured.

In fact I'm a bit divided on this, as I've rarely seen any real trouble from the trailing whitespace, except when people are strict on what to have. To me it just seems easier to edit when the blank lines are already indented to the correct position.

Many of the reasons in e.g. http://programmers.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal seem to apply only on trailing whitespace on lines with other content also, not to blank lines.

However, I have seen plenty of competent devs stripping out trailing whitespace, so I guess there is a point to it. If I may ask, could you explain-like-im-5-years-old, why is trailing ws so bad? :)

kylemanna:
Primary reason is to reduce false differences or deltas.  When updating the file all of my text editors reduce the white space on the adjacent lines.  To make the delta easier to follow, I prefer non-functional deltas be separated so I can concentrate more on the changes that change the operation of the code.

Also, for those who care about white space and have features like "set list" set on vim it will help them maintain sanity.

For me, It's a style preference and shows attention to detail which, infers quality when I read a new code base for the first time.

You're welcome to keep it however you like. :smile:

PetteriAimonen:
Fair enough. At least it will avoid nasty surprises for people whose editor strips the trailing ws :)

PR #181, opened 2015-10-15
Cmake

kylemanna:
* Enhancements that affect cmake only
* Build the protobuf generator files out of tree in the build directory
* Don't require `NANOPB_SRC_ROOT_FOLDER`


kylemanna:
Thanks!

PR #185, opened 2015-11-05
Add option to allow for anonymous unions

kamathba:
gnu99 and c11 both support anonymous unions, and being able to generate them could improve code quality in some cases.


kamathba:
Looks like I blew up macro expansions. This might take a few more looks.

PetteriAimonen:
Also, it is fine to only add support for anonymous unions now and leave anonymous structs for later, but in that case the comment in nanopb.proto should match.

PR #186, opened 2015-11-11
Add option to allow for anonymous unions

kamathba:
I have another implementation where I overload the 'anonymous' flag inside of the field rules as 'ONEOF' or 'ANONYMOUS_ONEOF'. I feel like I am relatively unfamiliar with a lot of the generator functionality so I would love feedback if you can think of a better way to do this, or at least one that is more consistent with the project style.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -518,8 +518,23 @@ struct pb_extension_s {
     pb_membersize(st, u.m[0]), 0, ptr}
 
 #define PB_ONEOF_FIELD(union_name, tag, type, rules, allocation, placement, message, field, prevfield, ptr) \
-        PB_ ## rules ## _ ## allocation(union_name, tag, message, field, \
+        PB_ ## ONEOF_ ## allocation(union_name, tag, message, field, \
         PB_DATAOFFSET_ ## placement(message, union_name.field, prevfield), \
+        PB_LTYPE_MAP_ ## type, ptr)
+
+#define PB_ANONYMOUS_ONEOF_STATIC(u, tag, st, m, fd, ltype, ptr) \
+    {tag, PB_ATYPE_STATIC | PB_HTYPE_ONEOF | ltype, \
+    fd, pb_delta(st, which_ ## u, m), \
+    pb_membersize(st, m), 0, ptr}
+
+#define PB_ANONYMOUS_ONEOF_POINTER(u, tag, st, m, fd, ltype, ptr) \
+    {tag, PB_ATYPE_POINTER | PB_HTYPE_ONEOF | ltype, \
+    fd, pb_delta(st, which_ ## u, m), \
+    pb_membersize(st, m[0]), 0, ptr}
+
+#define PB_ANONYMOUS_ONEOF_FIELD(union_name, tag, type, rules, allocation, placement, message, field, prevfield, ptr) \
+        PB_ ## ANONYMOUS_ONEOF_ ## allocation(union_name, tag, message, field, \
+        PB_DATAOFFSET_ ## placement(message, field, prevfield), \
         PB_LTYPE_MAP_ ## type, ptr)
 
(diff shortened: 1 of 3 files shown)


pb.h
```diff
@@ -518,10 +518,25 @@ struct pb_extension_s {
     pb_membersize(st, u.m[0]), 0, ptr}
 
 #define PB_ONEOF_FIELD(union_name, tag, type, rules, allocation, placement, message, field, prevfield, ptr) \
-        PB_ ## rules ## _ ## allocation(union_name, tag, message, field, \
```
kamathba: I also snuck in this change and forgot to revert it, but I think it is safely explicit in this case

PetteriAimonen:
Overall approach seems reasonable. I'll review more closely before merging.

The generator does not have any particularly smart overall design strategy, so anything that is reasonably clean and works is ok.

Did you happen to find any bugs during development that are not covered by current tests? (I guess nothing about anonymous unions is currently covered by tests, but I would like to add tests for anything that went wrong during your own manual testing.)

kamathba:
Hmmm my bugs mostly had to do with needing new macros and the 16/32-bit enum check. Things I thought about:

I thought the oneof.fields inheriting the anonymous property from the oneof object is a little convoluted, but didn't quickly think of anything better.

Does the generator check for name collisions? If so you would need some new logic to combine the anonymous union context with the context above it. Otherwise those types of errors will get caught in the compiler.

Similarly, having a field named which_*OriginalUnionName* would cause problems.

non GNU-C99 and earlier might have problems with anonymous unions and it would be good to catch that in pb.h, which might mean we need some pre-processor macros.

I can't think of any specific tests that an anonymous union-oneof implementation would need that a regular union implementation wouldn't. I'll sleep on it though.

PetteriAimonen:
Currently name collisions are not checked in generator, so they fall for the compiler to handle, which is fine IMO.

I think because this is not enabled by default, it is fine to fail in compilation stage if it does not support anonymous unions. Checking with preprocessor is difficult as many compilers support it as extension even if they do not support C99.

kamathba:
On a pure oneof test note, are oneofs inherently optional. AKA can the message that owns a oneof have NONE of the oneof tag/fields? is the behavior always the same if which_*union_name* is equal to an invalid tag (something that doesn't exist)

kamathba:
My assumption would be that nanopb wouldn't encode anything if the tag isn't valid, and would skip tags that it didn't have an understanding of.

PetteriAimonen:
Yeah, oneofs are always optional. This is indicated with which_oneof == 0 by the decoder, though I think encoder will handle any invalid tag the same.

pb.h
kamathba: Should probably either revert this or just make it PB_ONEOF_ ## allocation(...

pb.h
```diff
@@ -518,10 +518,25 @@ struct pb_extension_s {
...
+    fd, pb_delta(st, which_ ## u, m), \
+    pb_membersize(st, m), 0, ptr}
+
+#define PB_ANONYMOUS_ONEOF_POINTER(u, tag, st, m, fd, ltype, ptr) \
+    {tag, PB_ATYPE_POINTER | PB_HTYPE_ONEOF | ltype, \
+    fd, pb_delta(st, which_ ## u, m), \
+    pb_membersize(st, m[0]), 0, ptr}
+
+#define PB_ANONYMOUS_ONEOF_FIELD(union_name, tag, type, rules, allocation, placement, message, field, prevfield, ptr) \
+        PB_ ## ANONYMOUS_ONEOF_ ## allocation(union_name, tag, message, field, \
```
kamathba: same thing here

PetteriAimonen:
Thanks, merged this now. I did a few small fixes (biggest was a compilation error when oneof is not last field) and added a test case.

PR #187, opened 2015-11-19
Explicit python version in FindNanopb.cmake

danielotero:
CMake will use python2 binary instead of just python if the former
is found.

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -263,6 +263,7 @@ find_path(NANOPB_GENERATOR_SOURCE_DIR
 )
 mark_as_advanced(NANOPB_GENERATOR_SOURCE_DIR)
 
+set(Python_ADDITIONAL_VERSIONS 2)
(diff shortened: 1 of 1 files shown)


danielotero:
This fixes a issue in my system with both python2 and python3 binaries installed and /usr/bin/python link pointing to python3.

PetteriAimonen:
Hmm, with pull request #169 I think nanopb should work with python3
also now. I guess you don't have protobuf installed for python3?

danielotero:
You're right. My distribution only have the python2 protobuf version in the official repositories. I didn't find the python3 version and blamed Google too fast.

Sorry for bothering you.

PetteriAimonen:
No problem.

This might actually be a real issue, not sure if the CMake rules
should check for python3-protobuf and try python2 if not found.
Though if python3 is the default python for system, it makes sense
to expect that python3-protobuf would be the default protobuf
library.

PR #191, opened 2016-01-26
Fix incorrect operation on TI's DSC

baranovmv:
Some exotic platforms like [TI 28x](http://www.ti.com/lsds/ti/microcontrollers_16-bit_32-bit/c2000_performance/real-time_control/f2833x_f2837x/overview.page) aligns memory access to even addresses. This means that sizeof(char) equals sizeof(short) equals 1. This kind of platforms are not so rare as one may think.

This commit makes nanopb able to operate properly on such kind of platforms.

Here is brief description of changes:

**[u]int8_t** type lacks in stdint.h so I've changed it to **[u]int_least8_t**.

Also I've removed few dangerous address calculations like [that](https://github.com/baranovmv/nanopb/commit/8f7546b64c52d6bd48ce55dfa58dfece4371a600#diff-42b51cbb4326cbfca72e0b852032247fL447):
```
/* Here are access to different bytes from one integer value */
const uint8_t *bytes = value;
uint8_t lebytes[4];
lebytes[0] = bytes[3];
lebytes[1] = bytes[2];
lebytes[2] = bytes[1];
lebytes[3] = bytes[0];
return pb_write(stream, lebytes, 4);
```
and changed it to this:
```
const uint32_t *words = (const uint32_t*)value;
uint_least8_t bytes[4];
bytes[0] = (uint_least8_t)(*words >> 24) & 0xFF;
bytes[1] = (uint_least8_t)(*words >> 16) & 0xFF;
bytes[2] = (uint_least8_t)(*words >> 8) & 0xFF;
bytes[3] = (uint_least8_t)(*words) & 0xFF;
return pb_write(stream, bytes, 4);
```

This commit works on TMS320F28335 and intel x86 platforms.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -41,6 +41,10 @@
  * This was the default until nanopb-0.2.1. */
 /* #define PB_OLD_CALLBACK_STYLE */
 
+/* Switch this on if target platform provide unusual integer sizes.
+ * This takes places on platforms with minimal addreassable space bigger than
+ * one byte. Some TI's MCU or DSP for an example. */
+/* #define PB_UNUSUAL_INT_SIZE */
 
 /******************************************************************
  * You usually don't need to change anything below this line.     *
@@ -55,7 +59,7 @@
 /* Include all the system headers needed by nanopb. You will need the
  * definitions of the following:
  * - strlen, memcpy, memset functions
- * - [u]int8_t, [u]int16_t, [u]int32_t, [u]int64_t
+ * - [u]int_least8_t, [u]int_least16_t, [u]int32_t, [u]int64_t
  * - size_t
  * - bool
  *
@@ -144,7 +148,7 @@
  * Most-significant 4 bits specify repeated/required/packed etc.
  */
 
-typedef uint8_t pb_type_t;
+typedef uint_least8_t pb_type_t;
 
 /**** Field data types ****/
 
@@ -205,13 +209,13 @@ typedef uint8_t pb_type_t;
     typedef uint32_t pb_size_t;
     typedef int32_t pb_ssize_t;
 #elif defined(PB_FIELD_16BIT)
-#define PB_SIZE_MAX ((uint16_t)-1)
-    typedef uint16_t pb_size_t;
-    typedef int16_t pb_ssize_t;
+#define PB_SIZE_MAX ((uint_least16_t)-1)
+    typedef uint_least16_t pb_size_t;
+    typedef int_least16_t pb_ssize_t;
 #else
-#define PB_SIZE_MAX ((uint8_t)-1)
-    typedef uint8_t pb_size_t;
-    typedef int8_t pb_ssize_t;
+#define PB_SIZE_MAX ((uint_least8_t)-1)
+    typedef uint_least8_t pb_size_t;
+    typedef int_least8_t pb_ssize_t;
 #endif
 
 /* This structure is used in auto-generated constants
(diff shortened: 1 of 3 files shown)


PetteriAimonen:
The test fails seem to be due to some new warnings in GCC5. Not so important for now, they should be easy to fix.

More importantly, I'm not sure I want to be writing uint_least8_t everywhere. Maybe some typedef like pb_byte_t could be useful. Or just typedef uint_least8_t uint8_t in syshdr.h?

PetteriAimonen:
I changed pb_encode_fixedXX() functions so that they should work for your platform also.

Can you check whether everything else is fixed by:

    typedef uint_least8_t uint8_t;
    typedef uint_least16_t uint16_t;

placed in e.g. pb.h.

I know that it isn't very clean, but on the other hand uint_least8_t is pretty to annoying to type and use on platforms that do not care about the difference.

baranovmv:
>  the #ifdef is unnecessary

That's good point, the code looks more readable.

There is no [u]int8_t in stdint.h for TI 28xx. I'm doing ```typedef unsigned char uint8_t;``` in my own projects. So I can redefine ```PB_SYSTEM_HEADER``` in my projects. But I think it's better to avoid this problem for others. nanopb may redefine types in pb.h in this manner: ```pb_u8_t```, ```pb_s8_t``` etc. I think it would make nanopb more platform-agnostic.

Though the problem is that [this asserts](https://github.com/nanopb/nanopb/blob/master/pb.h#L248) fails in compiler:

```
Building file: ../log/log.pb.c
Invoking: C2000 Compiler
"/home/mkh/ti/ccsv6/tools/compiler/ti-cgt-c2000_6.4.4/bin/cl2000" -v28 -ml -mt --float_support=fpu32 --include_path="/home/mkh/ti/ccsv6/tools/compiler/ti-cgt-c2000_6.4.4/include" --include_path="/home/mkh/Coding/Axel/pb_exercises/log" -g --diag_warning=225 --display_error_number --diag_wrap=off --preproc_with_compile --preproc_dependency="log/log.pb.pp" --obj_directory="log"  "../log/log.pb.c"
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 253: error #95: the size of an array must be greater than zero
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 254: error #95: the size of an array must be greater than zero
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 255: error #95: the size of an array must be greater than zero
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 256: error #95: the size of an array must be greater than zero
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 257: error #95: the size of an array must be greater than zero
"/home/mkh/Coding/Axel/pb_exercises/log/pb.h", line 258: error #95: the size of an array must be greater than zero
```

This happens because sizeof(int16_t) == 1 on my platform. I propose to remove this asserts at all. Especially in case of pb's own type names.

PetteriAimonen:
Maybe the best compromise is to specify a new type, pb_byte_t, that will be used for byte buffers.
Making a pb_u8_t would in my opinion be equally bad as redefining uint8_t, as neither of these would really be 8 bits. Defining pb_byte_t on the other hand conveys just the semantics of storing a byte-sized piece of information.

The asserts are there for a reason, as things fail in interesting ways if someone messes up those typedefs. Some platforms fail to provide a suitable stdint.h, so it is quite common to copy-paste an unsuitable one. But the asserts could use CHAR_BITS to work better.

Do you know whether the C2000 compiler is freely available? What about a simulator for it? This being the first platform for nanopb where CHAR_BITS != 8, I might want to add it to the test runs.

baranovmv:
> Maybe the best compromise is to specify a new type, pb_byte_t, that will be used for byte buffers.
Making a pb_u8_t would in my opinion be equally bad as redefining uint8_t, as neither of these would really be 8 bits. Defining pb_byte_t on the other hand conveys just the semantics of storing a byte-sized piece of information.

Okay, sounds reasonable.

 > The asserts are there for a reason, as things fail in interesting ways if someone messes up those typedefs. Some platforms fail to provide a suitable stdint.h, so it is quite common to copy-paste an unsuitable one. But the asserts could use CHAR_BITS to work better.

If there is no stdint.h how one can get **CHAR_BITS**? May be these asserts should be placed under **PB_SYSTEM_HEADER** clause?

BTW there is definition of CHAR_BIT in limits.h of C2000 compiler:
```
#define CHAR_BIT                16    /* NUMBER OF BITS IN TYPE CHAR  */
```

> Do you know whether the C2000 compiler is freely available? What about a simulator for it? This being the first platform for nanopb where CHAR_BITS != 8, I might want to add it to the test runs.

One may use free license for [Code Composer Studio](http://www.ti.com/tool/ccstudio-C2000) I don't know about simulators for this platform.

PetteriAimonen:
How does it seem now?

I'm having a bit of trouble trying to figure any way to test this myself.

baranovmv:
It has compiled flawlessly,this is good sign. I'll test it tomorrow on a hardware.

Code looks clean and neat. Everything is fine, thank you!

PR #193, opened 2016-02-19
Add -D option to specify output directory

recursify:
As part of a build process, it's often nice to be able to specify where the output files go without having to move them after the fact.

Eventually it might be nice to have an option to output either the `.h` or `.c` file to stdout as well, so it can be piped anywhere.


PetteriAimonen:
Thanks!

PR #194, opened 2016-02-19
Generate MIN/MAX/ARRAYSIZE for enums

isotes:
This generates #defines mirroring the following values from the generated C++ code of GPB
* const Foo Foo_MIN: the smallest valid value of the enum (VALUE_A in the example).
* const Foo Foo_MAX: the largest valid value of the enum (VALUE_C in the example).
* const Foo Foo_ARRAYSIZE: always defined as Foo_MAX + 1.

The naming convention (e.g., `_MyEnum_MIN`) has been chosen to avoid conflicts with possible declared values.  `MyEnum__MIN` would be another obvious possibility.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -197,6 +197,10 @@ class Enum:
 
         result += ' %s;' % self.names
 
+        result += '\n#define _%s_MIN %s' % (self.names, self.values[0][0])
+        result += '\n#define _%s_MAX %s' % (self.names, self.values[-1][0])
+        result += '\n#define _%s_ARRAYSIZE ((%s)(%s+1))' % (self.names, self.names, self.values[-1][0])
+
         if not self.options.long_names:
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Patch seems fine, but just to make sure I understand the purpose, could you explain in what kind of situation would one use these defines?

Snaipe:
@PetteriAimonen not the OP, but this would help with maintainability when using enum values as indices in an array (e.g. map an enum value to a specific function pointer to handle a specific field)

isotes:
The primary use case for me is as @Snaipe described. Additionally, in the case of consecutive values for the enums (e.g., 0, 1, 2..., n), it also allows simple validity checks.

PetteriAimonen:
Ok, makes sense :)

Thanks for the patch!

isotes:
Thanks!

Issue #199, opened 2016-04-27
C naming conversion for struct and enum names


philips77:
Hi,
Right now, the generated code does not change the proto CamelCase into typical C naming style.
Is there an option to enabled it somewhere? Or do you plan to add this future?
If I have a single proto file and generate a code to C or Java it always look very ugly in one of the languages. Here https://developers.google.com/protocol-buffers/docs/style they say how fields should look like, but imho message names should be converted from Camel Case into C standard.

PetteriAimonen:
No, currently there is no such option and I haven't had plans to do this. But this seems like a reasonable feature request, though I'm not going to work on it myself in the near future.

M1cha:
additionally, there should be an option to remove the typedefs or remove the leading`_` from the structs so you can use `struct name` instead of `Name` as required by many codestyles.

PetteriAimonen:
This feature is now implemented thanks to @td-krzysiek!

I invite everyone to review the styling implemented.
If there are any changes that you think should be made, I prefer to do them before next stable release.

PR #200, opened 2016-04-29
Converting CamelCase to C style, adding _t for enums and messages

philips77:
The Protocol Buffers [guidlines](https://developers.google.com/protocol-buffers/docs/style) recommend using CamelCase for Enums, Messages and Services.
The current version of nanopb does not convert them to C style, but rather leaves some strange looking C code like:

```C
typedef struct _namespace_SomeMessage {
   namespace_Type some_value;
} namespace_SomeMessage;
```

This commit fixes this issue. It also uppercases the #define names for intt_defaults, tags, max size, and others.

The generated code look now like that:
```C
typedef enum
{
    NAMESPACE_TYPE_FIRST = 0,
    NAMESPACE_TYPE_SECOND = 1
} namespace_type_t;
#define NAMESPACE_TYPE_MIN NAMESPACE_TYPE_FIRST 
#define NAMESPACE_TYPE_MAX  NAMESPACE_TYPE_SECOND
#define NAMESPACE_TYPE_ARRAYSIZE ((namespace_type_t)(NAMESPACE_TYPE_SECOND+1))

typedef struct {
  namespace_type_t some_value;
/* @@protoc_insertion_point(struct:namespace_some_message_t) */
} namespace_some_message_t;

// etc.
#define NAMESPACE_MESSAGE_SIZE                            4
```

I've tested it with messages, enums, oneofs, extensions, required and optional, repeated, *max_length* and *max_size* defined in an options file, default values (numeric or enums). The C na H files compile every time without problems with pb.h version 30.

Example proto file:
```proto
package namespace;

enum Type {
  FIRST = 0;
  SECOND = 1;
}
message SomeMessage {
  required Type some_value = 1;
}
```

**Note:** The commit also removes the *_enum_name* and *_struct_name* from enums and structs definitions, but to add it again you can just uncomment it.
The commit also includes the hardcoded **_t** for enums and structs names.

This commit will break compatibility with previously generated code as type names and defines will change.

Also, I didn't try to build an .exe file from it.
Changes made by me and @hakonfam due to #199.

### DIFF
--- a/generator/camel_case_splitter.py
+++ b/generator/camel_case_splitter.py
@@ -1,0 +1,40 @@
+def split_camel_case(input):
+    def remove_camel_case(camel_case_input):
+        no_camel_case = ""
+        if len(camel_case_input) <= 0:
+            return ""
+        no_camel_case += camel_case_input[0].lower()
+        for c in camel_case_input[1:]:
+            if c.isupper():
+                no_camel_case += "_" + c.lower()
+            else:
+                no_camel_case += c
+        return no_camel_case
+
+    underscore_split = input.split("_")
+    retval = ""
+    should_be_upper = False
+    for i in underscore_split:
+        if i.isupper():
+            should_be_upper = True
+        if is_camel_case_name(i):
+            retval += remove_camel_case(i) + "_"
+        else:
+            retval += i + "_"
+
+    if should_be_upper:
+        retval = retval.upper()
+    return retval[:-1].replace("__", "_")
+
+
+def is_camel_case_name(input):
+    if '_' in input:
+        return False
+
+    if input.islower():
+        return False
+
+    if input.isupper():
+        return False
+
+    return True
--- a/generator/camel_case_splitter.py
+++ b/generator/camel_case_splitter.py
@@ -13,17 +13,12 @@ def split_camel_case(input):
 
     underscore_split = input.split("_")
     retval = ""
-    should_be_upper = False
     for i in underscore_split:
-        if i.isupper():
-            should_be_upper = True
         if is_camel_case_name(i):
             retval += remove_camel_case(i) + "_"
         else:
             retval += i + "_"
 
-    if should_be_upper:
-        retval = retval.upper()
     return retval[:-1].replace("__", "_")
 
 
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -458,7 +458,7 @@ class Field:
         if self.default is None:
             return None
 
-        ctype = self.ctype
+        ctype = self.ctype_t
         default = self.get_initializer(False, True)
         array_decl = ''
 
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -176,11 +176,11 @@ class Enum:
         self.names_upper = str(self.names).upper()
 
         if enum_options.long_names:
-            self.values = [(self.names + x.name, x.number) for x in desc.value]
+            self.values = [(str(self.names + x.name).upper(), x.number) for x in desc.value]
         else:
-            self.values = [(names + x.name, x.number) for x in desc.value]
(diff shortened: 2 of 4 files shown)


philips77:
One sec, I'll have a look at the bug.

bestouff:
Way nicer. Is that ok to break code compiled before your patch ?

Snaipe:
This should probably be an option at first, otherwise existing code will break.

Also, just a heads up, but this does not handle well acronyms: "GetHTTPMessage" for instance gets transformed to "get_h_t_t_p_message" instead of "get_http_message".

philips77:
Yes... And also I've just noticed that when the enum value in not written in capitals the generated name in small letters.
I need to spend more time on this one...

philips77:
Ok, now I see there is a bug in reading the options:

```
# Generate all fields as callbacks.
AllTypes.*              type:FT_CALLBACK
SubMessage.substuff1    max_size:16
AllTypes.oneof          no_unions:true
```
All fields were changed to pb_callback_t, not only those in AllTypes. But at least the first alltypes test passed :)
I will continue on Monday.

philips77:
It works when I change line 1457 to
```file_options = nanopb_pb2.NanoPBOptions()#get_nanopb_suboptions(fdesc, toplevel_options, Names([filename]))```
I'm not sure why but it read initial global options as type: FT_CALLBACK and applies type CALLBACK to *SubMessage.substuff1*, instead of STATIC.

PetteriAimonen:
Hmm, yeah, this should definitely be enabled with a separate option, and I think it should be off by default. I think it makes most sense to make it a .proto option, e.g. convert_camelcase = true.

Other than that, take your time :) I'll take a closer look once you are satisfied yourself, and I can of course help if you get stuck with some problem.

philips77:
I think the C style should be always enabled. The only thing that could be customizable is whether to add _t to enums and structs names, as POSIX doesn't have it. But using mixed Camel Case and C style in C looks very odd...

philips77:
Oh course there is a problem with applications that already use nanopb :) This is an argument to make it disabled by default.

PetteriAimonen:
No, the name style conversion cannot be enabled by default. It would break too much existing code for little benefit.. yeah.

bestouff:
Could it be enabled by default for proto3 ?

PetteriAimonen:
Possibly.. or even nanopb-0.4.0. Perhaps it is not such a burden to add one option for legacy code. But no breaking changes before 0.4.0, so it makes sense to keep it disabled at first.

philips77:
Hi, I've looked at the options in nanopb, trying to add one as you wrote. It looks like currently you have only per-field options. The new one (or two) should be kind of a global one? So I tried to add new message with global options to the nanopb.proto file, compile it.. but then things started to get very complicated with parsing. My Python experience is around 4 days long, including weekend, so I must give up :( 

We have now a version of the script that works for our need, but we hope someone, if not you, will continue with C-styling the generated files. I can't spend more time on it right now.

Snaipe:
Looking back, why would you add the POSIX-reserved `_t` suffix to the identifier of every type? This is just being non-conforming for the sake of being non-conforming.

philips77:
Hehe, yes. But this is actually a requirement in our C code standards. As I wrote, this should be also an option, default false. Of course you can just remove it completely, but it would be perfect for Nordic if such option could be enabled somehow without creating a private repo.

Snaipe:
This should probably be in a "suffix" option, imo. I've seen people use weird suffixes like `_s` as coding standard, so it should probably be configurable.

PetteriAimonen:
Yeah, perhaps it would make sense to have this more widely configurable. Something like:

string MessageNameTemplate = "msg_%s_t";
CaseType MessageNameCase = CamelCase;

etc.

PetteriAimonen:
Considering that the current pull-request is non-workable (breaks existing code and tests), I've moved this into feature request #217.

PR #204, opened 2016-06-06
Fix bad pointer dereference with static repeated submessages

CodyJung:

1. Have a repeated submessage inside a message, all of which are set to `type:FT_STATIC`
2. Call pb_release on the main message

**What happens?**
Read access violation when attempting to get the count of the submessage

**What should happen?**
Successful pb_release (basically a no-op for static fields, right?)

I believe this issue was introduced as a part of 88b2efe0477f4f9e313b5d7307dfd347b6893376. Let me know if this doesn't seem like the correct fix; I admit that I'm a little fuzzy about nanopb internals.

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -1034,13 +1034,13 @@ static void pb_release_single_field(const pb_field_iter_t *iter)
         
         if (PB_HTYPE(type) == PB_HTYPE_REPEATED)
         {
-            if(PB_ATYPE(type) == PB_ATYPE_POINTER)
-            {
-                count = *(pb_size_t*)iter->pSize;
-            }
-            else if(PB_ATYPE(type) == PB_ATYPE_STATIC)
+            if(PB_ATYPE(type) == PB_ATYPE_STATIC)
             {
                 count = iter->pos->array_size;
+            }
+            else
+            {
+                count = *(pb_size_t*)iter->pSize;
             }
         }
         
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -1034,7 +1034,14 @@ static void pb_release_single_field(const pb_field_iter_t *iter)
         
         if (PB_HTYPE(type) == PB_HTYPE_REPEATED)
         {
-            count = *(pb_size_t*)iter->pSize;
+            if(PB_ATYPE(type) == PB_ATYPE_POINTER)
+            {
+                count = *(pb_size_t*)iter->pSize;
+            }
+            else if(PB_ATYPE(type) == PB_ATYPE_STATIC)
+            {
+                count = iter->pos->array_size;
+            }
         }
         
         if (pItem)


PetteriAimonen:
Hmm.. can you provide code that reproduces the problem?

I couldn't get it to occur myself based on the description. What operations do you perform to the main message before you call pb_release()? E.g. how do you initialize the struct contents, or does it just have random values from memory?

If I understand the patch correctly, it changes field release so that static arrays would always release all entries of the array, disregarding the msg.field_count value. It is actually a bit questionable what would be better thing to do here. It initially makes more sense to me to rely on the _count value, ignoring the rest of the array entries. But due to idempotency of pb_release(), it should also be safe to do what your patch does, ignoring the _count value and releasing all array entries.

CodyJung:
Yeah! For reference, here's a (sanitized) version of the proto definitions we're using:
```
package int.Proto.Sample;

option optimize_for = LITE_RUNTIME;

message AEntry {
    optional uint32     one             = 1 [default = 0xFFFFFFFF];
    optional uint32     two             = 2 [default = 0xFFFFFFFF];
    optional uint32     three           = 3 [default = 0xFFFFFFFF];
    optional uint32     four            = 4 [default = 0xFFFFFFFF];
    optional uint32     five            = 5 [default = 0xFFFFFFFF, deprecated=true];
    optional uint32     six             = 6 [default = 0xFFFFFFFF];
    optional uint32     seven           = 7 [default = 0xFFFFFFFF];
}

message BEntry {
    optional uint32     uno             = 1 [default = 0xFFFFFFFF];
    optional uint32     dos             = 2 [default = 0xFFFF]; // uint16
    optional uint32     tres            = 3 [default = 0xFFFF]; // uint16
}

message CEntry {
    optional uint32     un              = 1 [default = 0xFFFFFFFF];
    optional uint32     deux            = 2 [default = 0xFF];   // uint8
}

message DEntry {
    optional uint32     eins            = 1 [default = 0xFFFFFFFF];
    optional float      zwei            = 2;
    optional float      drei            = 3;
    optional float      vier            = 4;
    optional float      funf            = 5;
    optional float      sechs           = 6;
}

message MainMessage {
    required bool                   sentinal                = 1;
    repeated AEntry                 a_entry                 = 2;
    repeated BEntry                 b_entry                 = 3;
    repeated CEntry                 c_entry                 = 4;
    repeated DEntry                 d_entry                 = 5;
}
```

And the options file:
```
* type:FT_STATIC
int.Proto.Sample.MainMessage.a_entry        max_count:7
int.Proto.Sample.MainMessage.b_entry        max_count:7
int.Proto.Sample.MainMessage.c_entry        max_count:7
int.Proto.Sample.MainMessage.d_entry        max_count:7
```

We also have the following three options enabled:
```c
#define PB_ENABLE_MALLOC 1
#define PB_FIELD_16BIT 1
#define PB_OLD_CALLBACK_STYLE
```

It _may_ be related to the fact that we're updating from nanopb-0.2.8 to 0.3.5. The issue occurs when we're done reading in a MainMessage encoded in the old version and are releasing it. We're not actually doing anything with it besides verifying that we can decode the message and the sentinel value is present (and in this case, we can and it is).

The message that the debugger is giving me is that dereferencing `iter->pSize` is failing; I'm not sure why pSize would be getting set to a bad pointer. I can keep looking, though.

PetteriAimonen:
Hmm, that is strange indeed. pSize (as in the pointer) should always be dereferenceable. Whether the value it points to is a valid size depends on more things, but having pSize itself be corrupt is already indicative of some other problem.

If you can get a dump of the pb_iter_t structure (so pos, pSize, pData etc.) when the crash happens, I could take a look.

CodyJung:
Something along these lines? This is the state of `iter` for each of the first three fields in MainMessage; it crashes on the third field.

```
//// ITER_BEGIN ////
    Iter
    ----
    start: 0x009af8f8 {WS.exe!const pb_field_s int_Proto_History_MainMessage_fields[6]}
    pos: 0x009af8f8 {WS.exe!const pb_field_s int_Proto_History_MainMessage_fields[6]}
    required_field_index: 0
    dest_struct: 0x28a8fe7c
    pData: 0x28a8fe7c
    pSize: 0x28a8fe7c //Note: *(pb_size_t*)0x28a8fe7c => 32744
    
    Iter.pos (0x009af8f8)
    ---------------------
    tag: 1
    type: 0 '\0'
    data_offset: 0
    size_offset: 0
    data_size: 4
    array_size: 0
    ptr: 0x00000000

//// ITER_NEXT ////
    Iter
    ----
    start: 0x009af8f8 {WS.exe!const pb_field_s int_Proto_History_MainMessage_fields[6]}
    pos: WS.exe!0x009af907 {tag=2 type=39 ''' data_offset=4 ...}
    required_field_index: 1
    dest_struct: 0x28a8fe7c
    pData: 0x28a8fe84
    pSize: 0x28a8fe80 //Note: *(pb_size_t*)0x28a8fe80 => 481

    Iter.pos (0x009af907)
    ---------------------
    tag: 2
    type: 39 '''
    data_offset: 4
    size_offset: -4
    data_size: 56
    array_size: 7
    ptr: 0x009af7e0 {WS.exe!const pb_field_s int_Proto_History_AEntry_fields[8]}

//// ITER_NEXT ////
    Iter
    ----
    start: 0x009af8f8 {WS.exe!const pb_field_s int_Proto_History_MainMessage_fields[6]} {tag=1 type=0 '\0' data_offset=...}
    pos: WS.exe!0x009af916 {tag=3 type=39 ''' data_offset=4 ...}
    required_field_index: 1
    dest_struct: 0x28a8fe7c
    pData: 0x28a90010
    pSize: 0x28a9000c //Note: *(pb_size_t*)0x28a9000c => unable to read memory

    Iter.pos (0x009af916)
    ---------------------
    tag: 3
    type: 39 '''
    data_offset: 4
    size_offset: -4
    data_size: 24
    array_size: 7
    ptr: 0x009af7a4 {WS.exe!const pb_field_s int_Proto_History_BEntry_fields[4]}
```

PetteriAimonen:
Yeah, that dump is useful.

pData: 0x28a90010
pSize: 0x28a9000c //Note: *(pb_size_t*)0x28a9000c => unable to read memory

This seems strange. The structure should be large enough that 0x28a9000c is well within it, and seems like the correct address for `MainMessage.b_entry_count`. Can you check what address the debugger gives for that `_count` field in the message structure?

(Perhaps also worth making sure that all your files have been recompiled after latest .proto change, as this seems a bit like the calling code is using different definition for the message structure.)

(As an unrelated issue, this got me looking at that array release code again from a new viewpoint and I noticed a bug #205. Shouldn't matter for your use case, though.)

CodyJung:
Wow, I feel really dumb. This was definitely just a pointer error on my part. Sorry for the false alarm.

I was doing:
```c
pb_release( int_Proto_Sample_MainMessage_fields, &decoded_pb_data );
```

instead of

```c
pb_release( int_Proto_Sample_MainMessage_fields, decoded_pb_data );
```

And it probably just worked by coincidence in 0.2.8. Bleh.

PetteriAimonen:
Nice to hear that you found the bug :)

PR #208, opened 2016-07-27
Added CMake project

podsvirov:
This will allow users to use the library from a package, rather than handling the implementation files.


PetteriAimonen:
A few questions:

1) Can you explain the difference between this and the "extra/FindNanopb.cmake" that we have currently? I'm a total cmake-novice myself.

2) Is it somehow possible to have the CMakeLists.txt in a subfolder instead of the root?

podsvirov:
Hello @PetteriAimonen,

work in progress and I'm ready for the debate.

> A few questions:

> 1) Can you explain the difference between this and the "extra/FindNanopb.cmake" that we have currently? I'm a total cmake-novice myself.

FindXXX.cmake is old school. I try create [relocatable package](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages).
This package can be installed in the development environment and used by everyone. 
Users do not need to download files FindXXX.сmake separately and add it to CMAKE_MODULE_PATH.

> 2) Is it somehow possible to have the CMakeLists.txt in a subfolder instead of the root?

No problems! We can place the *CMakeLists.txt* file in a *cmake* subdirectory, as in the original *protobuf* project.
Or do you have other preferences?

A one question:

What about rewrite generator in C or C++ language, because lite source C implementation required Python interpreter and environment?

PetteriAimonen:
Oh no, I have totally forgotten about this pull request. Sorry!
I'll merge it now.

As for generator being in Python; The generator only needs to run on the development PC. I provide binary versions of the generator, so it shouldn't be any different from an application written in C/C++ with a few libraries in that regard. And one would always need the protoc compiler and libprotobuf anyway, as I'm not going to reimplement .proto parsing.

podsvirov:
Hi! Monday began for me with the good news :-)
Thank you for your confidence. Thank you merged my changes.
I actively use CMake, and try to maintain these changes in the future.

PR #209, opened 2016-08-01
Add bazel BUILD file for nanopb.

mzhaom:


PetteriAimonen:
Is it possible to have the BUILD file in a subdirectory, or would that cause problems?

mzhaom:
It would be best to put it under the top directory. Is there any particular reason to put it in a subdirectory?

PetteriAimonen:
Mainly just that there are dozens of different build systems, so the top directory will easily be filled with files that are relevant to only a small number of people. But I guess that is not a big problem yet, I'll have to think about it later.

PR #211, opened 2016-08-03
Add inline allocation of bytes fields

tmroeder:
This commit adds a new FT_INLINE allocation type that forces bytes
fields to be inlined into the struct. E.g., pb_byte_t my_bytes[32].
Obviously, this requires max_size for the bytes field.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,24 @@
+Petteri Aimonen <jpa@npb.mail.kapsi.fi>
+Michael Poole <mdpoole@troilus.org>
+Daniel Kan <extremeblue99@gmail.com>
+Stan Hu <stanhu@aclimalabs.com>
+dch <david.hotham@blueyonder.co.uk>
+Steffen Siering <steffen siering gmail com>
+Jens Steinhauser <jens.steinhauser@gmail.com>
+Pavel Ilin <ilin.pa@gmail.com>
+Kent Ryhorchuk <kryhorchuk@xeralux.com>
+Martin Donath <scifish@gmail.com>
+Oliver Lee <oliverzlee@gmail.com>
+Michael Haberler <git@mah.priv.at>
+Nicolas Colomer <ncolomer@viadeoteam.com>
+Ivan Kravets <me@ikravets.com>
+Kyle Manna <kyle@kylemanna.com>
+Benjamin Kamath <ben.kamath@synapse.com>
+Andrew Ruder <andrew.ruder@elecsyscorp.com>
+Kenshi Kawaguchi <kenshi@recurse.ca>
+isotes <isotes@gmail.com>
+Maxim Khitrov <max@mxcrypt.com>
+Yaniv Mordekhay <yanivmo@users.noreply.github.com>
+Ming Zhao <mzhao@luminatewireless.com>
+Google, Inc.
+Tom Roeder <tmroeder@google.com>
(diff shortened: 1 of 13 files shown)


PetteriAimonen:
Is this only for fields that have constant length? Perhaps the decoder should probably check that the length matches and give an error if it doesn't.

I don't want to increase the size of the pb_field_t structure, as that affects the flash use on microcontrollers a lot. Perhaps "inline bytes fields" could instead be identified by new ltype, e.g. PB_LTYPE_FIXED_LENGTH_BYTES?

Also, is the change of the LICENSE.txt a mistake? Google is not directly involved in this project and does not own the copyright, even if some contributors may be Google employees (I am not).

tmroeder:
Thanks for the prompt feedback. I'll see about changing it to an ltype and making sure the fixed-length check is present.

I am following up with internal Google open-source folks, who were the ones that asked me to change the LICENSE file. I agree with your assessment, and I understand that removing it could be a condition of acceptance of the patch.

PetteriAimonen:
Sounds good.

I made a small error when reading the diff of the LICENSE file, thought it was replacing the line entirely instead of just adding the Google copyright. I agree that some projects add every contributor to the copyright list, but in the past I haven't done that in nanopb and I guess if one were to add Google, then I should probably list all the past contributors too (which are available from git log). But not sure if there is any benefit to list them there.

tmroeder:
For the contributors/LICENSE issue, would you accept the creation of an AUTHORS file that starts with you, then the Google, Inc line, then a note to see the git log for other authors and a note to these authors that they can submit a PR to get added to the file?

tmroeder:
or we could extract them from the git log, I suppose

tmroeder:
git log --all --format='%aN <%cE>' | sort -u | wc -l
26

That seems like a reasonable list, and it looks like there are a few duplicates with different email addresses. :)

tmroeder:
I've removed the line from LICENSE and added an AUTHORS file in the apparent git commit order. I squashed the changes into a single commit for a clean commit history.

I realized that I needed to add FT_INLINE and FT_LTYPE_FIXED_LENGTH_BYTES to the documentation, so I did that, too.

PetteriAimonen:
AUTHORS file looks great now :)
Thanks also for the typo fixes.

And, the code looks good also and I'll merge it now :)

mxk:
I there a reason for limiting this to fixed-length fields? I need an option that will compile this:

```protobuf
message Msg {
	bytes data = 1;
}
```
to this:
```c
typedef struct _Msg {
	pb_size_t data_size;
	pb_byte_t *data;
} Msg;
```

FT_POINTER gives you a pointer to pb_bytes_array_t, which forces the size and data to be contiguous. That's an inconvenient limitation for encoding, since I want to pull the data from elsewhere without copying it first. Am I right in assuming that I have to use callbacks to achieve this?

PetteriAimonen:
@mxk Yeah, currently that needs callbacks. These forum posts are relevant:
https://groups.google.com/forum/?fromgroups#!searchin/nanopb/bytes$20pointer/nanopb/JydwGfCC-YA/8CPXECBUxjEJ
https://groups.google.com/forum/?fromgroups#!searchin/nanopb/bytes$20pointer/nanopb/pAXyfjk5E_Q/Z9-kirlCs_MJ

It's admittedly not a perfect situation, but generating the size field separately had its own problems, especially when having an array of bytes fields.

PetteriAimonen:
I extended the fixed-length bytes field to support proto3 and also pointer type fields in issue #244. Should be fully backwards compatible, but there is also now a separate option `(nanopb).fixed_length = true` which is preferred for new use over `FT_INLINE`.

PR #215, opened 2016-09-15
RFC: Add "implicit has" option for optional messages

fazjaxton:
I am using nanopb in an embedded project for storage of nonvolatile data to flash.  In this application, I would like for the users of the data to be able to assume that values in nonvol-backed structs are current and valid, and that changes they make to the structs will be automatically backed up to nonvol at shutdown.  I do not want to push the responsibility of checking and setting a lot of "has_xxx" fields onto the users of this data.

I am using protobuf for the ease of encoding and decoding this data, especially where fields can be added and removed from version to version of firmware.  When a new field is added, it gets loaded to its default value, and all the old nonvol values are retained.  All fields in my protobuf structs are "optional", so that you can upgrade and downgrade versions without trouble.

This change adds an "optional_implicit_has" nanopb option.  For any field with this set to "true", it does not generate the "has_" field, and always assumes that the field is present when encoding.  When decoding, if the field is not present, it is set to default like usual.  Since we must have enough flash to store all values, the larger encoding size of "implicit_has" does not make a difference.

This is mostly a proof-of-concept.  I am glad to change anything related to coding style, naming, etc.  Please let me know what you think, and if this is a valuable addition to upstream.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -237,6 +237,7 @@ class Field:
         self.default = None
         self.max_size = None
         self.max_count = None
+        self.implicit_has = False
         self.array_decl = ""
         self.enc_size = None
         self.ctype = None
@@ -255,6 +256,9 @@ class Field:
 
         if desc.HasField('default_value'):
             self.default = desc.default_value
+
+        if field_options.HasField("optional_implicit_has"):
+            self.implicit_has = field_options.optional_implicit_has
 
         # Check field rules, i.e. required/optional/repeated.
         can_be_static = True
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Hmm, I'd rather see something that is closer to what protobuf version 3 does.
In proto3 the rules are:
- All fields are optional
- Field is present whenever its value is non-zero

I think that is not far from what you have already implemented.
What would you think about doing these changes:
- Instead of new "implicit_has" field in the Python script, set field.rules = "SINGULAR" (this is the term that proto3 uses for its optional fields).
- In encoder, when size field is missing, check if the value is non-zero before encoding it. (This will require a bit of logic based on each field type https://developers.google.com/protocol-buffers/docs/proto3#default, I can also do it later in a separate commit if it seems too troublesome now.)
- Call the field option "proto3" or similar. This will then later be set automatically if the file is protobuf v3, but nothing stops setting it manually for use in protobuf v2.

fazjaxton:
Thanks @PetteriAimonen!  I agree, and think I can implement all three of those.  I'll update and resubmit.

berni155:
@fazjaxton Are you still working on these points? If not I could try to work on the suggested solutions from @PetteriAimonen within the next week.

PetteriAimonen:
I've merged #216 now.

fazjaxton:
@berni155 - Sorry for the delay ... I've been away from e-mail.  Thanks for getting this put in.  :)

PR #216, opened 2016-10-09
Proto3 singular fields support

berni155:
Based on Pull request #215 and the comments of @PetteriAimonen. Thanks to @fazjaxton


PetteriAimonen:
Looks good and I'll merge this now.

It will need a testcase, but I think I'll add that myself (so that I also get to try out the new feature ;)

PR #223, opened 2016-11-21
nanopb: update generator to emit optional enum->string mapping function

wak-google:
We need an enum to string method for being able to trivially print enum encoded error codes.


PetteriAimonen:
Merged with a small change: I renamed the generated function from MyEnum_Name() to MyEnum_name() to match with the other generated identifiers (_fields etc.)

wak-google:
Thanks

PR #224, opened 2016-11-21
Make pb_field_t void* member pointer size aligned

wak-google:
For some reason the linker will zero-out these pointer values at
link time if this is not aligned properly to store a 64bit pointer on powerpc.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -241,7 +241,7 @@ struct pb_field_s {
     /* Field definitions for submessage
      * OR default value for all other non-array, non-callback types
      * If null, then field will zeroed. */
-    const void *ptr;
+    const void *ptr __attribute__((aligned(sizeof(void*))));
 } pb_packed;
 PB_PACKED_STRUCT_END
 


PetteriAimonen:
Does defining "PB_NO_PACKED_STRUCTS" help? That could be what is causing it not to be aligned by default.

wak-google:
Yes, but that doesn't explain why the compiler / linker is doing weird stuff and zeroing out the pointer. Digging deeper to figure out what the correct fix is here.

wak-google:
The difference in the struct just from the above change can be seen with objdump:

Aligned:
```
Contents of section .rodata:
 a8210 00000008 27000000 04ffffff fc000000  ....'...........
 a8220 18000000 03000000 00000000 000a80d0  ................
```

Unaligned:
```
Contents of section .rodata:
 a81e9 000000 08270000 0004ffff fffc0000 00 ....'...........
 a81f9 180000 00030000 00000000 0000        .............
```

wak-google:
We are fine not using packed struct so we can drop this.

PR #226, opened 2016-12-09
Fix potential unaligned access

kurddt:
Hello,

**What happens ?**
When trying to use nanopb on ARMv4 I came across a trap when encoding a message. I figured out that an unaligned access cause this. 

**How to reproduce the problem**
This occur only when you have a string of `size == sizeof(uint64_t)` and the data pointer is not aligned to a 64bit boundary.

If the type is string, do not try to deference it as int16, int32 or int64.
This may lead to unalign memory access, which may cause trap on some architectures (ARM)

### DIFF
--- a/pb_encode.c
+++ b/pb_encode.c
@@ -220,11 +220,15 @@ static bool checkreturn encode_basic_field(pb_ostream_t *stream,
             if(bytes->size == 0)
                 implicit_has = false;
         }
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks, I'll merge your fix soon; also added issue #227 because this might get hit by someone else also.

kurddt:
Your welcome.

PR #230, opened 2016-12-23
Fix closing a non-empty substream resulting in an incorrect stream state

Tobba:
If pb_close_string_substream is called on a non-empty substream bytes_left will end up being decreased by more than what was actually read. This commit checks and empties out the substream first if it's non-empty.

Not a problem unless you're calling the substream functions yourself since nanopb itself always exhausts the stream before closing, but probably not the intended behaviour.

### DIFF
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -11,6 +11,19 @@ are included, in order to make it easier to find this document.
 
 .. contents ::
 
+Nanopb-0.3.8 (2017-xx-xx)
+=========================
+Fully drain substreams before closing
+
+**Rationale:** If the substream functions were called directly and the caller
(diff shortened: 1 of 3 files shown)


Tobba:
Ah crap, I forgot about checking for errors on the buf_read. Doing so would need to change pb_close_string_substream to be able to error; which would be breaking. Comments?

PetteriAimonen:
Changing pb_close_string_substream() to return bool is small enough breaking change that it can be done any time. Just have to add note to docs/migration.rst

Tobba:
Fixed up & squashed.

PetteriAimonen:
Everything looks good, thanks! :)

PR #236, opened 2017-01-17
cmake: Fix library name

wak-google:
The produced static library should be `libprotobuf-nanopb.a`
instead of the current `liblibprotobuf-nanopb.a`.

Signed-off-by: William A. Kennington III <wak@google.com>


PetteriAimonen:
Thanks :)

PR #241, opened 2017-02-07
cmake: Clean up the install heirarchy and install the generator

wak-google:
This is a fairly extensive change that removes some of the unecessary
vendoring that is done and changes the namespaces for the nanopb options
proto. It cleans up the cmake build so that nanopb can be installed into
a linux userspace, and have programs depend on it.

This also breaks out the configuration options from pb.h into a separate
config.h which is not provided by default. If you use the cmake build
system this will be automatically provided with the most compatible
settings.

Signed-off-by: William A. Kennington III <wak@google.com>


PetteriAimonen:
Hmm, there are fairly lot of breaking changes.

For example, is there a specific need for changing nanopb_msgopt to nanopb.msgopt? This seems like it will just cause a lot of unnecessary work for the current users.

PetteriAimonen:
(I do agree that your way of organizing nanopb.proto might be cleaner, but the difference is so minor that it is not worth it to change it at this point of the library development.)

wak-google:
The biggest problem is that when installing the proto / headers without this change, it doesn't result in placing them inside of a package directory. This is fine for c with the typical include directory structure. Unfortunately for python the package should have it's own directory. You can work around this by passing arguments to protoc during generation time to correctly include from the right package path. Unfortunately that needs to be maintained by any dependent projects.

PetteriAimonen:
Hmm, yeah, that makes sense. However the namespace change is so big that it will have to wait for  0.4.0.

Are there any parts that could be merged separately, or does it make more sense to do it all at once?

PetteriAimonen:
Now that I think about it, the new and old options format should be binary compatible in the protobuf format. So would there be any problem if the old nanopb.proto is left in place so that old .proto files can include it, while new software could use the new options.proto?

wak-google:
I think that's perfectly fine to wait for a major release to include a
change like this. I figured that's what would happen anyway. It might even
make sense to have a separate branch for major changes going into the next
major release that might break user code. Anything like this would work for
me. I can also add an entry somewhere for notable breaking changes if you
would like.

The old protos should be binary compatible with the new one since none of
the message tags changed, so we could still keep the old one around in the
new release. Let me know what you think makes the most sense.

wak-google:
Any updates on your end? Would you like to proceed with these changes in a new major release / branch?

PetteriAimonen:
Yeah, here's the plan:

1) I'll add the new options.proto in parallel with old nanopb.proto to 0.3.8 and release that next week
2) I'll start a 4.0 dev branch after that, the rest of changes can go there.

PetteriAimonen:
(Don't mind the conflicts, I'll take care of them when I merge)

CMakeLists.txt
```diff
@@ -30,35 +30,95 @@ if(NOT DEFINED CMAKE_INSTALL_CMAKEDIR)
...
+        else()
+            set(protoc_FLAGS "-I/usr/include")
+        endif()
+    endif()
+    message(STATUS "Protoc Flags: " ${protoc_FLAGS})
 
     foreach(generator_proto IN LISTS generator_protos)
-        string(REGEX REPLACE "([^;]+)" "generator/proto/\\1.proto" generator_proto_file "${generator_proto}")
-        string(REGEX REPLACE "([^;]+)" "\\1_pb2.py" generator_proto_py_file "${generator_proto}")
+        string(REGEX REPLACE "([^;]+)" "${CMAKE_SOURCE_DIR}/generator/nanopb/\\1.proto" generator_proto_file "${generator_proto}")
```
conz27: ${PROJECT_SOURCE_DIR} is a better option. If you have a top-level CMakeLists.txt that includes this one in a add_subdirectory() call, this logic will break because the resulting path will be rooted at the top-level CMakeList's folder.

CMakeLists.txt
```diff
@@ -30,35 +30,95 @@ if(NOT DEFINED CMAKE_INSTALL_CMAKEDIR)
...
+        string(REGEX REPLACE "([^;]+)" "${CMAKE_CURRENT_BINARY_DIR}/\\1_pb2.py" generator_proto_py_file "${generator_proto}")
```
conz27: ${PROJECT_BINARY_DIR} == ${CMAKE_CURRENT_BINARY_DIR} in this case, but it's probably better to be explicit about it rather than rely on CMAKE_CURRENT_BINARY_DIR.

CMakeLists.txt
```diff
@@ -30,35 +30,95 @@ if(NOT DEFINED CMAKE_INSTALL_CMAKEDIR)
...
         add_custom_command(
             OUTPUT ${generator_proto_py_file}
-            COMMAND protoc --python_out=${CMAKE_CURRENT_BINARY_DIR} -Igenerator/proto ${generator_proto_file}
+            COMMAND protoc ${protoc_FLAGS}
+                --python_out=${CMAKE_CURRENT_BINARY_DIR}
+                -I"${CMAKE_SOURCE_DIR}/generator/nanopb" ${generator_proto_file}
```
conz27: ${PROJECT_SOURCE_DIR} same as reason above.

CMakeLists.txt
```diff
@@ -30,35 +30,95 @@ if(NOT DEFINED CMAKE_INSTALL_CMAKEDIR)
...
     endforeach()
+
+    install(
+        PROGRAMS generator/nanopb_generator.py generator/protoc-gen-nanopb
+        DESTINATION ${CMAKE_INSTALL_BINDIR}
+    )
 endif()
 
 if(nanopb_BUILD_RUNTIME)
+    include_directories(BEFORE ${CMAKE_CURRENT_BINARY_DIR})
```
conz27: ${PROJECT_BINARY_DIR}

CMakeLists.txt
```diff
@@ -30,35 +30,95 @@ if(NOT DEFINED CMAKE_INSTALL_CMAKEDIR)
...
+
     add_library(protobuf-nanopb STATIC
+        ${CMAKE_CURRENT_BINARY_DIR}/config.h
```
conz27: ${PROJECT_BINARY_DIR}

CMakeLists.txt
```diff
@@ -85,6 +149,16 @@ if(nanopb_BUILD_RUNTIME)
         ${CMAKE_CURRENT_BINARY_DIR}/nanopb-config-version.cmake
```
conz27: ${PROJECT_BINARY_DIR}

conz27:
Sorry for the spam ... I wanted to capture it all in one comment, but it didn't work the way I thought it would.

wak-google:
@conz27 Fixed

pedrokiefer:
Will this be merged to 0.4.0 dev branch?

I've started to add full support for bazel, using the master branch, and I think this clean up would help a lot. The only thing I see is missing, would be to move the source files from the repo root to a src or nanopb directory, then I would not have any problem with the includes.

PetteriAimonen:
@pedrokiefer Which part of this clean up is most important to you?

I'm still having a lot of doubts about changing the nanopb.proto into its own namespace, but the rest should be mergeable.

Cloudef:
Having the ability to install generator systemwide would be great. (As well as the .pc files)

PetteriAimonen:
I'm sorry for hanging onto this for so long. But this pull request is too large for me to understand, and some parts of it I've decided are not worth the effort. Especially changing `nanopb.proto` namespace is something I'm currently not willing to do.

If anybody has something specific here that they'd like to see in nanopb main branch, please try to separate it into smaller issues / pull requests.

PR #245, opened 2017-02-25
Fix infinite loop in pb_check_proto3_default_value

jdemar:
- Occurs with proto3, PB_FIELD_16BIT and submessage > 255 bytes
- Add test case for PB_FIELD_16BIT and proto3


PetteriAimonen:
Hmm, indeed. I wonder if checking the submessage for all-zero byte values is the wrong solution after all, and instead it should recurse into caling `pb_check_proto3_default_value` for every field in submessage. I'll look into the matter.

PetteriAimonen:
Thanks, this at least fixes the immediate problem. I'll see about changing the comparison method later.

PR #250, opened 2017-03-05
CMAKE: remove use of relative paths

conz27:
When doing out of source builds:

mkdir build
cd build && cmake -G "Unix Makefiles" ../ && make

The build script tripped up on relative paths of /generator/proto and
the files under generator/proto/*.proto.

By using ${PROJECT_SOURCE_DIR}, the paths become absolute and the issue
disappears.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,12 +46,12 @@ if(nanopb_BUILD_GENERATOR)
         string(REGEX REPLACE "([^;]+)" "\\1_pb2.py" generator_proto_py_file "${generator_proto}")
         add_custom_command(
             OUTPUT ${generator_proto_py_file}
-            COMMAND protoc --python_out=${CMAKE_CURRENT_BINARY_DIR} -I${PROJECT_SOURCE_DIR}/generator/proto ${generator_proto_file}
+            COMMAND protoc --python_out=${PROJECT_BINARY_DIR} -I${PROJECT_SOURCE_DIR}/generator/proto ${generator_proto_file}
             DEPENDS ${generator_proto_file}
         )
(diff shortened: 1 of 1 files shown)


conz27:
This issue affects 0.3.8 and nanopb-0.3.8.

PetteriAimonen:
@wak-google What do you think about this? I must admit that I don't really use cmake myself so I'm not that well equipped to know how the script should work.

conz27:
Some context around this change:

I am using nanopb is a git submodule to my project, and I have a CMakeList.txt file for my entire project that is responsible for also compiling and installing nanopb among other things. 

Since the build & install commands are triggered from a folder higher up the tree than the nanopb repo folder, the relative paths used by nanopb's CMakeLists.txt become incorrect because Makefile generated by CMake thinks generator/proto folder exists on the level where the build was triggered from instead of in ${nanopb}/generator/proto.

CMake automatically defines ${PROJECT_SOURCE_DIR}  and ${PROJECT_BINARY_DIR} variables  because of the line "project(nanopb)" in CMakeLists.txt. The changes I made then correctly transform the relative paths into absolutely paths by prepending by PROJECT_* variables above in front of the relative paths for nanopb and plugin files.

conz27:
@wak-google , can you please take a look at this when you have chance? Would like to get it in before 0.4.0. It's a subset of your change #241, which I noticed after putting up this pull request.

wak-google:
Looks good to me

conz27:
@PetteriAimonen ... I think this pull request is ready for merge?

PetteriAimonen:
Yeah, thanks for the reminder. I've been a bit busy so forgot.

PR #252, opened 2017-03-17
Speed up deocding fixed width integers

haraldF:
Remove some unneeded bit-shifting by reading directly to an aligned
buffer.

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -1140,34 +1140,49 @@ bool pb_decode_svarint(pb_istream_t *stream, int64_t *dest)
 
 bool pb_decode_fixed32(pb_istream_t *stream, void *dest)
 {
-    pb_byte_t bytes[4];
+    union {
+        uint32_t fixed32;
+        pb_byte_t bytes[4];
+    } u;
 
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
This look like it will not work for big-endian platforms, nor for platforms that do not have 8-bit datatypes.

I think most compilers should be able to optimize away the shifting. Which comiler are you using and with what options?

haraldF:
Fixed the case for big-endian platforms.

I don't think the compiler can optimize away the shfits as the byte buffer is potentially unaligned. With the union, the buffer is forced to align to 32/64 bits, so we can use an assignment rather than a byte-per-byte copy on little endian machines. But I haven't verified the generated assembly code, I should do that :)

I'm currently using clang building for WASM32 backend, which is a bit experimental.

PetteriAimonen:
Hmm yeah, looks like GCC on ARM doesn't optimize it very well either.

I guess with a CHAR_BIT check added to the #if this could go in. It will need also adding #include for limits.h in pb.h, but that should be fine.

PetteriAimonen:
Thanks, this is now merged in branch dev_0.4.

PR #257, opened 2017-05-01
The max size for encoding strings must leave spaces for '\0'

Ahris:
The max size for encoding strings must leave spaces for a null terminator in the case where the field type isn't a pointer. This issue caused a critical bug because the proto was able to encode, but couldn't then decode without the null terminator.


PetteriAimonen:
Hmm yeah, I think this makes sense.

I'll see what the test failure is about, and will merge this for 0.4 branch. It is possible that someone relies on being able to encode up to the maximum size, so I don't want to break things in 0.3.9.

Ahris:
Woot! Updated the unit tests for str enc to match the code change and they're passing now. 

Thanks. :-)

PetteriAimonen:
Thanks, this is now merged in branch dev_0.4.

PR #265, opened 2017-06-28
In extra/FindNanopb.cmake, add dependency for .options files

kfitch:
In extra/FindNanopb.cmake, make the generated .pb.c and. pb.h files depend on the .options file (if it exists), so that editing the .options file will trigger a rebuild.


PetteriAimonen:
Thanks!

PR #170, opened 2015-09-21
Add Travis CI Support

kylemanna:
Travis CI will provide a simple automated mechanism to run the unit tests to make sure every pull requests works as expected on a matrix of compilers and python releases.

This is my first stab at making it work.  My repo's builds are @ https://travis-ci.org/kylemanna/nanopb

A Travis CI account will need to be setup for the main nanopb repository.

Related to #169

PR #283, opened 2017-09-15
Fix Travis-CI Build

kylemanna:
This fixes the recent Travis-CI breakage discussed in issue nanopb/nanopb#170.

The fix is to update to newer packages and directly invoke scons.  Seems that upstream has fixed somethings.

The build is now twice as fast thanks to using pre-built `protoc` binary provided by the Github release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,11 +42,17 @@ before_install:
 #    directories:
 #        - $HOME/protobuf
 
+# Rather then compile protobuf 3 from source, use the binaries now available
+# to speed up build time and reduce surprises until Ubuntu adds protobuf3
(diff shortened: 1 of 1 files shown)


kylemanna:
For some reason this [PR didn't trigger a Travis-CI build](https://travis-ci.org/nanopb/nanopb/pull_requests)?  Was it disabled?

For reference, my build is at https://travis-ci.org/kylemanna/nanopb

PetteriAimonen:
Strange about it not triggering, let's see what happens after merge..

PetteriAimonen:
Great, it works now! Thanks a lot!

PR #284, opened 2017-09-18
Added pb_decode_delimited_noinit

elcojacobs:
This seemed to be missing form the API.
It's simply pb_decode_delimited, but without initializing the target structure. Similar to pb_decode vs pb_decode_noinit.

This is useful for partial updates to an existing data structure, which is how I use it myself.

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -1017,6 +1017,21 @@ bool checkreturn pb_decode(pb_istream_t *stream, const pb_field_t fields[], void
     return status;
 }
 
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Hmm, yeah. I guess it makes sense to have it.

Starts to be quite many decode_* variants, maybe I'll convert some of them as arguments in 0.4.0.

elcojacobs:
Thanks! I think it especially makes sense in this case. Because when you are receiving partial updates and want to merge, you don't know what's in the message and need the delimited version of the command.

About the function name vs arguments, I would care mostly about code size. Whatever compiles to smaller code gets my preference.

PR #286, opened 2017-09-25
Use absolute path for NANOPB_OPTIONS_FILE in CMake

charlesincharge:
As far as I can tell, the NANOPB_GENERATOR_EXECUTABLE in https://github.com/charlesincharge/nanopb/blob/2769d8cc22c6f1c00a198dd11d81b81b726c82ba/extra/FindNanopb.cmake#L231 is run out of the build directory and not the source directory. Since the .options file is not copied to the build directory, the absolute path to the .options file should be used, rather than the path relative to the source directory.

Issue reproduction steps, checkout https://github.com/charlesincharge/nanopb/tree/excluded-cmake-options-file and build the cmake_simple example. Based on the .options file, the generated simple.pb.h should define `lucky_bytes` as a PB_BYTES_ARRAY_T instead of a pb_callback_t.


PetteriAimonen:
@kfitch What do you think? Should the .options file be copied to the build directory or should it be done with the absolute path?

charlesincharge:
It probably makes more sense for the .options file to be copied to the build directory. @PetteriAimonen, do you want me to change this PR to do that?

PetteriAimonen:
Yeah, that would be nice. I don't know much about CMake myself, but that's how I think it should be done.

PetteriAimonen:
I think this has been fixed by commit f4a588dcdca7b in pull request #312.

Issue #285, opened 2017-09-25
PB_WITHOUT_64BIT and negative numbers


smorin-primero:
When using the PB_WITHOUT_64BIT defines, if we encode a negative number in a var int, the value is not compatible with the value generated when the define is not used.

It probably has to do with the fact that negative number are normally encoded on 64 bit.

I was working on implementing something similar, and I had to make a special case for negative value to be encoded:

`
bool checkreturn pb_encode_negative_varint(pb_ostream_t *stream, uint32_t value)
{
  pb_byte_t buffer[10];
  size_t i = 0;
  size_t compensation = 32;/* we need to compensate 32 bits all set to 1 */

  while (value)
  {
    buffer[i] = (pb_byte_t)((value & 0x7F) | 0x80);
    value >>= 7;
    if (compensation)
    {
      /* re-set all the compensation bits we can or need */
      size_t bits = compensation > 7 ? 7 : compensation;
      value ^= ((0xFF >> (8 - bits)) << 25); /* set the number of bits needed on the lowest of the most significant 7 bits */
      compensation -= bits;
    }
    i++;
  }
  buffer[i - 1] &= 0x7F; /* Unset top bit on last byte */

  return pb_write(stream, buffer, i);
}
`

PetteriAimonen:
Hmm yeah. The issue #97 raises its head again.

PR #290, opened 2017-10-05
Fix negative number without 64bit

smorin-primero:
This is a fix for #285, I modified the unit test to take this into account.


PetteriAimonen:
Thanks!

PR #298, opened 2017-11-27
Fix #296 - install path of Python plugins with CMake.

### LINKED ISSUE #296: Install path of Python plugin incorrect when using relative CMAKE_INSTALL_PREFIX
Hi there, I've noticed that if I do an out-of-tree build with CMake, and then configure CMAKE_INSTALL_PREFIX to be "../", the nanopb library and include files go to the correct location, but the python plugins end up in "../../".  Similarly, if I specify CMAKE_INSTALL_PREFIX="../../", the Python plugins end up in "../../../../" (as long as I have write permission.

To reproduce, start with a clean 0.3.9 tree.
cd nanopb
mkdir build
cmake ..
ccmake .. (edit the CMAKE_INSTALL_PATH to be "..")
[configure, generate]
make
make install

Note it seems to work, if you configure with "cmake .. -DCMAKE_PREFIX_PATH=..", since it turns it into an absolute path first.

dflogeras:
No need of specifying the CMAKE_PREFIX_PATH to python when asking for
sysconfig.get_python_lib since everything is installed relative to
CMAKE_PREFIX_PATH anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,7 @@ if(nanopb_BUILD_GENERATOR)
     find_package(PythonInterp 2.7 REQUIRED)
     execute_process(
         COMMAND ${PYTHON_EXECUTABLE} -c
-            "from distutils import sysconfig; print(sysconfig.get_python_lib(prefix='${CMAKE_INSTALL_PREFIX}'))"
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks!

PR #299, opened 2017-12-14
Require non-buggy version of cmake

ezyang:
See https://github.com/pytorch/pytorch/issues/4141 and https://github.com/gflags/gflags/issues/113 for context


PetteriAimonen:
Seems reasonable to me.

Pinging @dflogeras, @wak-google in case you want to take look.

PR #304, opened 2018-01-08
Add nanopb_PROTOC_PATH variable to CMakeLists.txt

rjshade:
Allows use of a protoc binary in a non-standard location (e.g. a repository that includes a specific checked-in version of protoc).


PetteriAimonen:
Thanks.

PR #305, opened 2018-01-09
Explicitly handle all enum values in switch

rjshade:
Building nanopb with clang via Xcode 8.1 fails with [-Werror, -Wswitch-enum]. This PR explicitly handles the missing enum value in the switch statement.

nanopb/pb_decode.c:334:13: error: enumeration value 'PB_WT_STRING' not explicitly handled in switch [-Werror,-Wswitch-enum]
    switch (wire_type)
            ^
1 error generated.


PetteriAimonen:
Thanks!

Issue #260, opened 2017-05-30
Introduce fixed_count option for fixed sized arrays


jimon:
Currently it's only possible to set `max_count` for `repeated` fields, would be nice to have `fixed_count` for cases when we know exactly how many items we will have. This will allow to save space on `_count` field in the struct.

girtsf:
+1. I would find this useful as well.

PetteriAimonen:
Yeah, seems reasonable. I'll just copy my notes from the forum here so I'll find them later:

FIXED_LENGTH_BYTES is now generated as its own LTYPE. The obvious way
for FIXED_COUNT_REPEATED would be to have its own HTYPE, but there are
no unused values there. But I think that one could just have it be a
repeated field where pb_field_t.size_offset == 0, and then make the
decoder & encoder take the count from pb_field_t.array_size.

PetteriAimonen:
Reopening until fix is released.

jimon:
@PetteriAimonen wow, great! thanks a lot!

PetteriAimonen:
@jimon Most of the thanks really belongs to @oliverlee :)

PR #309, opened 2018-02-13
Add fixed count option

oliverlee:
Fixes https://github.com/nanopb/nanopb/issues/260


PetteriAimonen:
Hmm, the storage of the array count for non-packed fields is indeed problematic. Modifying iter->pData is not a good idea because that can mess up the decoding of following fields. Theoretically multiple packed arrays can occur in overlapping orderings if protobuf messages are merged by appending.

E.g.
repeated int32 A = 1 [(nanopb).fixed_count = true];
repeated int32 B = 2 [(nanopb).fixed_count = true];

the entries for A and B could occur like A1, B1, A2, B2, A3, B3 and it would be a valid protobuf message. Not very common but that merging is a useful property that some users of protobuf rely on.

Not sure how to solve that problem, seems like it would require some extra storage in the struct to be able to keep track of it.

oliverlee:
Ah okay. I haven't read the protobuf spec so I wasn't aware of that.

How about only allowing fixed_count for packable fields for now?

PetteriAimonen:
That could work. It's should also be ok to accept non-packed arrays when the count matches the fixed_count, and give error message when it doesn't. At least this way it will be easy enough to debug if someone happens to use the two features (fixed_count and message appending) in combination.

Maybe the cleanest way to do it would be to declare `const pb_field_t *fixed_count_field = NULL; pb_size_t fixed_count_size = 0;` variables in `pb_decode_noinit()` outside the loop. Then initialize `pSize` around line 955 like:

<pre>
if (PB_HTYPE(iter.type) == PB_HTYPE_REPEATED && iter.pSize == iter.pData)
{
    if (fixed_count_field != iter.pos)
    {
        /* Start decoding fixed size array, use auxiliary variable to count entries */
        if (fixed_count_field != NULL && field_count_size != fixed_count_field->array_size)
        {
            PB_RETURN_ERROR(stream, "wrong size for fixed count array");
        }
       fixed_count_field = iter.pos;
       fixed_count_size = 0;
    }
    iter.pSize = &fixed_count_size;
}
</pre>

And that `PB_RETURN_ERROR` if also has to be duplicated after the loop to check the count for the last fixed_count field.

This should handle a single fixed count array even when it is unpacked and unordered among several fields, because the count is only reset at the beginning of the next fixed count field. And it should report errors if there is wrong count for the field.

Other things: 
- The test set reports that the generator gives error `AttributeError: ExtensionRange instance has no attribute 'fixed_count'`. You can just add initializer for that field to ExtensionRange constructor.
- This feature will require some test cases. `tests/inline/` could act as a template, it is the basic test case for fixed_length bytes (dating from back when it used to be called "inline bytes"). To make a new testcase, just copy the folder to a new name and rename the files. You can run test cases with `scons -u` inside the test case folder. If making tests seems too complex I can of course do that myself also.
- If you can remove commit 3df84 (simple example modification) from the pull request, that will make it easier to merge. While a reasonable demo of this feature, I don't think it should go into the simple example in the main repo.

PetteriAimonen:
Thanks, I'll take a closer look later this week.

Thanks also for the scons python 3 fix, I hadn't even noticed that scons now supports Python 3 also. It probably picks the version the scons module was installed with.

The CMake fixed are also welcome but it can be clearer if they are in separate pull request.

oliverlee:
I submitted another PR with the CMake fix.

Thanks for your help!

PetteriAimonen:
Looks good, thanks for the doc updates also :)

PR #311, opened 2018-02-15
Always return size for messages with OneOf fields

oliverlee:
Use a maximum calculation if submessage sizes are unknown.


PetteriAimonen:
Looks good, thanks!

Issue #294, opened 2017-10-31
Publish Conan package


shreyasbharath:
If you haven't heard of [Conan](https://www.conan.io/), it's a C/C++ package manager that is gaining a lot of popularity.

I think it'd be great if every release of Nano Protocol Buffers was automatically published as a new version of a Conan package. It'll make consuming it in other projects really easy, and also updating to newer versions.

PetteriAimonen:
Ok, can you perhaps make a pull request about what would be needed to achieve this?

shreyasbharath:
Sure, I'll create a PR when ready.

PR #313, opened 2018-02-21
WIP: #294 - Create conan package recipe

shreyasbharath:
This is a first stab at implementing #294


shreyasbharath:
This is a first stab at implementing https://github.com/nanopb/nanopb/issues/294

BUILD
```diff
@@ -8,14 +8,14 @@ cc_library(
   name = "nanopb",
   visibility = ["//visibility:public"],
   hdrs = [
-    "pb.h",
-    "pb_common.h",
-    "pb_decode.h",
-    "pb_encode.h",
+    "src/pb.h",
```
shreyasbharath: It looks like I broke a bunch of things by moving these files.

Can you give me some direction as to how to fix them?

PetteriAimonen:
Well, I've thought about moving those files myself a few times. But moving them will break a lot of projects that rely on nanopb, for a quite small benefit. So I don't think it is worth moving them.

As for pushing to conan-center, there will have to be some kind of automation for that. Otherwise it'll be just too bothersome to do.

shreyasbharath:
Okay will move those files back to where they were.

We should be able to make Travis CI automatically deploy to conan-center every time you tag a new release. I will look into this.

bmulder-innoseis:
Is there still work ongoing at this? I tried to build it myself, but it wants python-protobuf installed, which again depends on protoc being installed system-wide. Installing python-protobuf with a hardcoded path to protoc still did not solve this issue.
So what's the status?

shreyasbharath:
Hi, apologies for the late reply. I have't been able to make any more progress on this. Let me spend a half hour and see if I can get the PR up to scratch.

PetteriAimonen:
A build recipe for conan is still welcome, but I'm closing off this pull request because it's not mergeable as-is. Especially the file moving is something I'm currently not willing to do.

PR #314, opened 2018-02-23
Updates descriptor.proto to the latest.

AndrewBallinger:
Many options have been added to descriptor.proto since the last time it
was pulled in and nanopb will not work on .proto files using the new
options.

Addressed more permanently by https://github.com/nanopb/nanopb/pull/241
but that PR makes a series of larger changes as well.


PetteriAimonen:
Thanks!

PR #315, opened 2018-02-27
Bytes init value fix

jipsuh:
Fixes init values for byte types. Protoc uses octal format for default values that aren't your typical characters, which was handled incorrectly by nanopb. For example protoc turns "\xff" to "\377" which in turn is encoded character by character by nanopb (taking 4 bytes of space). This was changed in protoc at some point, nanopb 0.3.4 (iirc) worked fine.

There may still be some corner cases that aren't handled properly. Certain special characters and such. More exhaustive test cases would be good to be added.

Also did some minor changes to scons files to add python 3 support (New scons versions run on python 3 as well as 2).

Juha-Pekka Hamina
R&D engineer
Nordic Semiconductor

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -442,32 +442,32 @@ class Field:
         else:
             return []
 
-    def to_hex_array(self, value):
-        retval = []
+    def to_hex_array(self, default_values):
+        '''Converts a default value bytestring to a list of hex values in string format
+           handles possible octal escapes inserted by protoc'''
+        hex_array = []
         i = 0
-        first = False
-        while i < len(value):
+        while i < len(default_values):
             # Check for escaping
-            if(value[i] == "\\"):
+            if(default_values[i] == "\\"):
                 # If the value is an escaped backslash...
-                if(value[i + 1] == "\\"):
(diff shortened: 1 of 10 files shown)


PetteriAimonen:
Hmm, would it be possible to use Python's own decoding function, e.g. `self.default.decode('string_escape')`?

Looks like the documentation has remained the same, so nanopb should have always de-escaped this; but later protoc versions escape more characters.

jipsuh:
I tried out some of Python's own functions but ran into issues such as:
`    decoded = self.default.decode('string-escape')`
`AttributeError: 'str' object has no attribute 'decode'`

Still I think using a library function to do the decoding would be better than the self made approach that I took. I believe that with some digging it should be possible to find one. Although there might not be a single function that works both under Python 2 and 3.

Feel free to take over and refine on this subject. I was originally tasked to only upgrade to a newer nanopb version on our systems and having to debug and fix caused some unexpected delay for me.

PetteriAimonen:
Ok, indeed there might be python 2 vs. python 3 incompatibilities hiding there.

This is ready to merge as-is, I can clean it up later if I feel like it. Thanks for the fix!

PR #320, opened 2018-04-05
Cmake relpath fix

KidVizious:
FindNanopb.cmake is broken when using the relative path feature.  This breakage was introduced in commit e64b3348978b4f273d21033ff1f7dcb5f50ec15c which switched from using the two-step compilation method to using nanopb's plugin mode.  

Path's are handled differently between these two modes and this change caused the generated *.pb.h and *.pb.c files to be placed in an incorrectly nested directory hierarchy.  This can be verified by running the cmake_relpath example project.


PetteriAimonen:
Hmm yeah, handling relative paths through protoc is problematic because it doesn't always give the full path info to plugins.

I'm very much a CMake novice, @oliverlee what do you think about this fix?

oliverlee:
That was an oversight on my part. The output directory should be `${CMAKE_CURRENT_BINARY_DIR}/${FIL_PATH_REL}` in the event that the relative path is not the current binary directory.

KidVizious:
@oliverlee The problem that I have found is that the protoc compiler includes the relative path, so compiling a file like sub/test.proto will be written to sub/test.pb.*.  If you add the relative path to the binary dir like ${CMAKE_CURRENT_BINARY_DIR}/${FIL_PATH_REL}, then the file gets written to sub/sub/test.pb.*.

oliverlee:
Yeah you're right. I didn't realize the relative path was already included and should have run the test.

PetteriAimonen:
So hmm, I assume this is fine to merge then? :)

PR #333, opened 2018-08-28
Feature: Flattening nested protobuf names

matejcik:
In our codebase, we use protobuf `package`s and message nesting to organize code. It's great, but it means that in the resulting C code, we would have to deal with _insanely long_ identifiers such as `hw_trezor_messages_bitcoin_TxAck_TransactionType_TxInputType`. In our case, the nesting and organization is useful in other places, but not in the C codebase - it turns out that we can use a flat structure.

This patch introduces a feature that allows us to do two things:

1. strip local package name from type names - useful when you don't have cross-package references
2. strip all but the last component of the type name - useful in cases like ours, when the nested identifiers are unique and you don't need the nesting in C code.

I'm not sure if you're interested in a feature like this, or if this is a good way to implement it. Worth a shot though :)

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -111,7 +111,7 @@ class Names:
     def __add__(self, other):
         if isinstance(other, strtypes):
             return Names(self.parts + (other,))
-        elif isinstance(other, Name):
+        elif isinstance(other, Names):
             return Names(self.parts + other.parts)
         elif isinstance(other, tuple):
             return Names(self.parts + other)
@@ -190,10 +190,13 @@ class Enum:
         self.options = enum_options
         self.names = names
 
+        # by definition, `names` include this enum's name
+        base_name = Names(names.parts[:-1])
+
         if enum_options.long_names:
             self.values = [(names + x.name, x.number) for x in desc.value]
         else:
-            self.values = [(desc.name + x.name, x.number) for x in desc.value]
+            self.values = [(base_name + x.name, x.number) for x in desc.value]
 
         self.value_longnames = [self.names + x.name for x in desc.value]
         self.packed = enum_options.packed_enum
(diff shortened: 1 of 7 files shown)


PetteriAimonen:
Seems like an useful feature and the implementation looks ok also.

This needs a testcase though. If it feels difficult, I can of course add one myself also.

matejcik:
I can try to write one tomorrow, but your test system confuses me. Could you point me to a existing testcase on which this could be based? Or do you have a testing guide?

PetteriAimonen:
I agree that it can be a bit confusing - if you can get the test cases to run by typing `scons` in the tests folder, that's a good beginning.

I think a simple testcase that has a package name and the option and references messages by the short name could cover the basics - i.e. it wouldn't compile if the generated name was wrong.

Though on the other hand, there is already the `tests/package_name` test case, which tests that the package name is included in identifiers - one could modify this to tests that it is not included, and it would automatically test the `AllTypes` structure that has all different field types etc. covered.

If you feel like it, try it out. But indeed there is no guide and I can do it myself it ends up difficult.

jlubawy:
+1 for this feature

How does this work with options? Would you refer to a field by its full-name or short-name if you flattened the name?

For example, would it be:

```
hw_trezor_messages_bitcoin_TxAck_TransactionType_TxInputType     enum_to_string:true
```

or 

```
TxInputType     enum_to_string:true
```

matejcik:
That's up for debate. In the current implementation, I'm using shortened names - it makes a little more sense to me to have the `options` match the result, not the source.

But you could also argue that options should be independent of the generated source, i.e., use original names.

matejcik:
made a testcase and it looks like it's working

PetteriAimonen:
Looks great, thanks a lot!

PR #340, opened 2018-09-18
Add --fast option to generator

vlovich:
Generate explicit encode/decode options for each protobuf.  When --fast is provided the codegen within those functions is altered so to be optimized for performance.  Rather than using descriptors the codegen explicitly generates code for each type.  This trades off more code size usage for less usage of constants and faster encode/decode.  This can be a more appropriate tradeoff on some embedded platforms (if there's limitations on amount of const usage or performance is more critical
than absolute smallest code footprint).

Without --fast the generated encode/decode functions can still be used and are just fully inlined pass-through wrappers to pb_encode/pb_decode.

Rather than using the generic pb_encode/pb_decode routines, there's new:
- decode_X
- get_encoded_size_X
- encode_X
routines where X is the protobuf message name.  Without --fast these simply redirect to the generic pb_encode/pb_decode routines with 0 overhead (code or performance) assuming a decent inlining compiler.  With --fast these routines are auto-converted to the speed-optimized version which
lets you benchmark before & after really easily.

The has_ fields are also optimized to generate as bitfields rather than individual bytes.  In addition to the speed benefits these changes also:
- reduces stack pressure when putting protobuf structs on the stack
- reduces heap pressure when putting protobuf structs on the heap
- reduces stack pressure because call-depth of speed-optimized encode/decode is much shallower (no recursion)

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -253,6 +253,22 @@ class Enum:
 
         return result
 
+    def is_valid_definition(self, static):
+        result = ''
+        if static:
+          result += 'static '
+        result += 'bool %s_IsValid(uint32_t varint) {\n' % self.names
+        result += '  switch (varint) {\n'
+        for _, number in self.values:
+          result += '    case %d:\n' % number
+        result += '      return true;\n'
+        result += '    default:\n'
+        result += '      return false;\n'
+        result += '  }\n'
+        result += '}\n'
+        return result
+
+
 class FieldMaxSize:
     def __init__(self, worst = 0, checks = [], field_name = 'undefined'):
         if isinstance(worst, list):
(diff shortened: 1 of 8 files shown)


PetteriAimonen:
Hmm, I don't particularly like the idea of spitting out C code directly from the generator. It makes it harder to edit the C code, and also more difficult to apply security updates.

Instead I think a better approach could be to have the relevant parts of code as some kind of macro constructs that would get expanded by C compiler. Kind of like how the PB_FIELD list in .pb.c files currently works. Perhaps it could even reuse the same definition of the fields.

But apart from that, do you have benchmarks on how much speedup this actually provides?

vlovich:
I don't have benchmarks for M4 yet (in the process of gathering it).  On CSR8675 this was ~50-100+ times faster.  Code size increase wasn't significant and is project-specific.  We had ~100 protos which translated to ~35kb more code using a *very* old version of GCC (& CSR8675 tends to bloat more quickly than traditional ARM code IIRC).  More importantly the CSR8675 has actually much less const space (which is made worse by 1 byte = 16bit); without this optimized version we had significant pressure on our const space from protobufs.

I'm not sure about how the macro approach could work.  Please take a look at the generated code first to confirm that it's possible to use macros but my initial thought is that it's not really feasible (at least not in any meaningful way that would actually make code maintenance easier).

vlovich:
Here are the numbers I've collected so far on an M4:

Empty message - 3x faster encode, 1.86x faster decode
Small message - 1.4x faster encode, 1.07x faster decode
Large message - 1.12x faster encode, 1.19x faster decode
Many large messages:
Complex fields messages:

Empty message: no definition
Small message: a small array
Large message: a large array
Many fields message: 10 fields, 8 of which are arrays
Complex fields message: lots of nesting etc

```

message MessageComplexBenchmarkProtobuf {
   message Nested2 {
       required uint32 field1 = 1;
       optional uint32 field2 = 2;
       oneof oneof {
           string field3 = 3;
           bytes field4 = 4;
       }
   }

   message Nested1 {
       required uint32 field1 = 1;
       optional uint32 field2 = 2;
       optional Nested2 field3 = 3;
   }

   oneof oneof {
       uint32 field1 = 1;
       uint32 field2 = 2;
       Nested1 field3 = 3;
       Nested2 field4 = 4;
   }
   Nested1 field5 = 5;
   Nested2 field6 = 6;
}
```

PetteriAimonen:
Thanks for the benchmarks. This is something that I'll perhaps look into in the future. But the approach of generating code directly from python generator seems like a too large maintenance burden, and for that reason I'm closing this pull request for now.

PR #345, opened 2018-10-10
pb_decode double free error with garbage input, PB_ENABLE_MALLOC and callbacks

oseiskar:
Hello. We've encoutered a bug and this PR contains a failing test demonstrating the issue:

If decode callbacks are used together with ([untrusted](http://jpa.kapsi.fi/nanopb/docs/security.html#division-of-trusted-and-untrusted-data)) garbage data and PB_ENABLE_MALLOC, a double-free occurs somewhere inside `pb_release`, called from `pb_decode` causing a segmentation fault.

This does not happen without PB_ENABLE_MALLOC or without decode callbacks (the test in the first commit of this PR passes)

`scons` output (tested on Debian Jessie)
```
... snip ...
[ OK ]   All patterns found in build/inline/inline.pb.h
run_test(["build/mem_release/mem_release.output"], ["build/mem_release/mem_release"])
Command line: ['build/mem_release/mem_release']
mem_release: build/common/malloc_wrappers.c:45: counting_free: Assertion `alloc_count > 0' failed.
[FAIL]   Program build/mem_release/mem_release returned -6
scons: *** [build/mem_release/mem_release.output] Error -6
scons: building terminated because of errors.
```

Thanks for taking a look!

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -1158,7 +1158,7 @@ static void pb_release_single_field(const pb_field_iter_t *iter)
             ext = ext->next;
         }
     }
-    else if (PB_LTYPE(type) == PB_LTYPE_SUBMESSAGE)
+    else if (PB_LTYPE(type) == PB_LTYPE_SUBMESSAGE && PB_ATYPE(type) != PB_ATYPE_CALLBACK)
     {
         /* Release fields in submessage or submsg array */
         void *pItem = iter->pData;
--- a/tests/mem_release/mem_release.c
+++ b/tests/mem_release/mem_release.c
@@ -177,9 +177,30 @@ static bool test_OneofMessage()
     return true;
 }
 
+/* Garbage input */
+static bool test_Garbage()
+{
+    const uint8_t buffer[] = "I'm only happy when it rains";
+    const size_t msgsize = sizeof(buffer);
+
+    {
+        OneofMessage msg = OneofMessage_init_zero;
+        pb_istream_t stream = pb_istream_from_buffer(buffer, msgsize);
+        TEST(!pb_decode(&stream, OneofMessage_fields, &msg));
+    }
+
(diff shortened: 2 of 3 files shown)


oseiskar:
also ping @Cloudef

pb_decode.c
```diff
@@ -1184,8 +1184,8 @@ static void pb_release_single_field(const pb_field_iter_t *iter)
                 count = iter->pos->array_size;
             }
         }
-        
-        if (pItem)
+
+        if (pItem && PB_ATYPE(type) != PB_ATYPE_CALLBACK)
```
PetteriAimonen: This check could perhaps be moved up to line 1161, because the lines in between have no effect for callbacks.

Cloudef: I first tried to put this check as precondition for the whole function `if (PB_ATYPE(type) != PB_ATYPE_CALLBACK) return;`, as how I reasoned, releasing a callback field doesn't ever seem to make sense. But it seems that caused some tests to fail, so was hoping you could give better comment where to do this check properly.

PetteriAimonen: Yeah. It's annoyingly delicate to get all the cases right here. Extension fields have "CALLBACK" type, but they contain a pointer to a field, which may be a submessage, which may contain pointer fields. That's why it would fail tests if you return at the beginning of the function.

Cloudef: Do you think handling this on case 1161, would be the best way then?

PetteriAimonen: Yeah, I think it is. I can of course change that myself also, I'm just not currently at a machine where I can run the test set.

Cloudef:
Cleaned up the commits, and changed the location of the callback check.

PetteriAimonen:
Thanks, merged now.

PR #358, opened 2018-11-01
Clarify the usage of callback fields inside oneof sections in the docs

### LINKED ISSUE #357: Does nanopb support submessages with callback fields in an onof field?
The issue is about nanopb 3.9.1.

I know nanopb does not support first-level callback fields in oneof, like the following
```
message Msg {
   oneof data {
        string s = 1;
        int32 i = 2;
    }
};
```
Without setting the maximum size of `s`, the generator will report an error like the following

> Callback fields inside of oneof are not supported (field s)

However, the generator will no complain if the message definition is revised to
```
message Wrapper {
    string s = 1;
};

message Msg {
    oneof data {
        Wrapper w = 1;
        int32 i = 2;
    }
};
```

Even though, the decoding of the wrapped string is not actually successful in my test. 

I've looked at nanopb_generator.py. It seems that the script does not recursively check if any nested fields are of the callback type. Is this intended? If not, I can make a patch to the generator and make it complain about nested callback fields.

uraj:
Add a few lines to the docs to resolve #357.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -317,6 +317,11 @@ The user is expected to set the filed manually using the correct field tag::
 Notice that neither ``which_payload`` field nor the unused fileds in ``payload``
 will consume any space in the resulting encoded message.
 
+When a C union is used to represent a ``oneof`` section, the union cannot have
+callback fields or nested callback fields. Otherwise, the decoding process may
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks

PR #365, opened 2018-11-23
Remove plugin.proto from CMakeLists.txt

thestr4ng3r:
Follow-up to https://github.com/nanopb/nanopb/commit/a10c771f1e1ae21f3afcce88ce128c80c0b28ec6 to fix cmake


PetteriAimonen:
Thanks, didn't catch that in my grep.

PR #366, opened 2018-11-30
Append NO_CMAKE_FIND_ROOT_PATH to find_file and find_path

a1ien:
I encountered the same as taka-no-me/android-cmake#95 problem when using buildroot.
Another more info link https://reviews.llvm.org/D41622
Can we add NO_CMAKE_FIND_ROOT_PATH to find_file and find_path?

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -292,6 +292,7 @@ endif()
 find_path(NANOPB_INCLUDE_DIRS
     pb.h
     PATHS ${NANOPB_SRC_ROOT_FOLDER}
+    NO_CMAKE_FIND_ROOT_PATH
 )
 mark_as_advanced(NANOPB_INCLUDE_DIRS)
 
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Hmm, that's beyond my cmake knowledge.

@oliverlee, @twam, @kfitch What do you think?

oliverlee:
It seems unlikely that you would want to search in CMAKE_FIND_ROOT_PATH, especially since PATHS are provided in the calls to find_x(). I guess this is probably fine.

@a1ien Are you cross-compiling and did you set CMAKE_FIND_ROOT_PATH_MODE_INCLUDE to ONLY? If so, you can use [find_host_package()](https://github.com/taka-no-me/android-cmake#finding-packages) or define an equivalent [macro](https://github.com/taka-no-me/android-cmake/blob/556cc14296c226f753a3778d99d8b60778b7df4f/android.toolchain.cmake#L1535-L1553).

a1ien:
@oliverlee thanks about  find_host_package.
This is not me CMAKE_FIND_ROOT_PATH_MODE_INCLUDE to ONLY set by [buildroot/toolchainfile.cmake](https://github.com/maximeh/buildroot/blob/master/support/misc/toolchainfile.cmake.in). I'm write package file for application that use nanopb.
Whit this patch all builds good.

PetteriAimonen:
Thanks, and sorry for the delay in merging.

PR #370, opened 2018-12-28
Correction of WIN32 flag in generator function

MaderPL:
Previously used flag (WIN32) indicated target system on which compiler
was running. Proposed change (CMAKE_HOST_WIN32) will indicate if host
system is windows.

This allows to cross compile for windows on linux machine or cross
compile for linux on windows machine. Also this change won't affect
builds for native linux or native windows.


PetteriAimonen:
Thanks!

PR #378, opened 2019-02-11
Implement conan recipe

### LINKED ISSUE #294: Publish Conan package

amatosov-rbx:
Fixes #294

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,0 +1,6 @@
+from conan.packager import ConanMultiPackager
+
+if __name__ == "__main__":
+    builder = ConanMultiPackager(build_policy="outdated")
+    builder.add_common_builds(shared_option_name=None)
+    builder.run()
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
How should `version = "0.3.9.2"` be handled?

It seems like it should be 0.4.0-dev in the master now, and `tools/set_version.sh` should be changed to update also that recipe.

amatosov-rbx:
good point. I'll fix that

PetteriAimonen:
(The failing check is just due to travis having some download troubles again.)

amatosov-rbx:
I've pushed version patching in `set_version.sh`, tested it with gnu-sed on macOS and it patches all files properly.

PetteriAimonen:
Thanks!

PR #384, opened 2019-03-01
Add C++ message descriptors helper

vlovich:
Zero-overhead. Allows creating simpler general nanopb C++ wrappers by
examining variable type and not needing macro magic.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1042,6 +1042,16 @@ class Message:
 
         return result
 
+    def fields_declaration_cpp_lookup(self):
+        result = 'template <>\n'
+        result += 'struct MessageDescriptor<%s> {\n' % (self.name)
+        result += '    static PB_INLINE_CONSTEXPR const pb_size_t fields_array_length = %d;\n' % (self.count_all_fields())
+        result += '    static inline const pb_msgdesc_t* fields() {\n'
+        result += '        return &%s_msg;\n' % (self.name)
+        result += '    }\n'
+        result += '};'
+        return result
+
     def fields_definition(self, dependencies):
         '''Return the field descriptor definition that goes in .pb.c file.'''
         width = self.required_descriptor_width(dependencies)
@@ -1454,6 +1464,18 @@ class ProtoFile:
         yield '} /* extern "C" */\n'
         yield '#endif\n'
 
+        if options.cpp_descriptors:
+            yield '\n'
+            yield '#ifdef __cplusplus\n'
+            yield '/* Message descriptors for nanopb */\n'
+            yield 'namespace nanopb {\n'
+            for msg in self.messages:
+                yield msg.fields_declaration_cpp_lookup() + '\n'
+            yield '}  // namespace nanopb\n'
+            yield '\n'
+            yield '#endif  /* __cplusplus */\n'
+            yield '\n'
+
         # End of header
         yield '/* @@protoc_insertion_point(eof) */\n'
         yield '\n#endif\n'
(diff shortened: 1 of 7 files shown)


PetteriAimonen:
Hmm, perhaps this should go behind some option. Either on compiler command line, or a message-specific option.

I've been considering moving other rarely used stuff like protoc insertion points behind separate options also, because while they're zero overhead when compiled, they clutter the generates files quite a bit.

PetteriAimonen:
(The travis check fail is also valid: // is not a valid comment in C89, you have to use /* */)

vlovich:
Got it. Will hide it behind a command. Reorganizing can be deferred once you've figured out how you want to organize that stuff?

PetteriAimonen:
Yeah. It wouldn't be any problem for me to add an option for it later, but as this is a new feature it might make sense to have it behind an option to start with.

Otherwise it looks mergeable. I still kind of wish there was a native C++ protobuf library for embedded, but in the absence small helpers like this are useful :)

vlovich:
Well nanopb is a great spot for that :D. Do you have any pointers on how I can write some unit tests for this?

PetteriAimonen:
Well, you can use .cxx files in unittests quite the same way as .c files.
Something quite basic should suffice, i.e. just access the template with the name it should have and check that code compiles and the values returned are correct. Doesn't even have to encode or decode anything. The "enum_to_string" testcase might be a reasonable template, it also tests just a few simple functions.

As for enabling the option, use NANOPBFLAGS variable in SConscript: https://github.com/nanopb/nanopb/blob/master/tests/regression/issue_203/SConscript

vlovich:
That doesn't seem to work. The test runs with both environments appended to each other

vlovich:
--nanopb_out=--cpp-descriptor--strip-path:.

vlovich:
NVM. Figured it out.

vlovich:
OK. I made some more changes vs the last thing you reviewed:

1. Hide the descriptor behind namespace nanopb
2. Add support for C++ versions between C++03 & C++17
3. Add tests for C++03, C++11, C++14, & C++17 versions of the generated code (maybe there's a cleaner way you'd like me to do it?
4. Fix regression 203 SConscript to not adjust the global env.

PetteriAimonen:
Hmm yeah, looks good otherwise than the duplication of code in the tests.

Perhaps it could be done in single SConscript like this:

     env_cpp03 = env.Clone()
     env_cpp03.Append(CXXFLAGS = '-std=c++11')
     p = env_cpp03.Program('test_cpp03', ['message_descriptor.cc', 'message.pb.c'])
     env_cpp03.Run(p)
     
     env_cpp11 = env.Clone()
     env_cpp11.Append(CXXFLAGS = '-std=c++11')
     p = env_cpp03.Program('test_cpp11', ['message_descriptor.cc', 'message.pb.c'])
     env_cpp11.Run(p)

So that the source code doesn't need to be duplicated. Easier to add extra checks later.

vlovich:
Sounds good.

vlovich:
Not familiar enough with scons. How do I modify my PR to skip if the compiler doesn't support the flag?

I tried
```
    e.Append(CXXFLAGS = '-std={}'.format(std))

    conf = Configure(e)
    if not conf.CheckCXX():
        print("Skipping {} test - compiler doesn't support it".format(std))
        continue
```

but then SCons just exits with an error but no message about what I'm doing wrong:
```
Checking whether the C++ compiler works... yes

scons: *** 
```

vlovich:
Hacked something together that works - not sure if that's the cleanest way to write it. Let me know.

PetteriAimonen:
Thanks, merged!

PR #411, opened 2019-07-12
plugin include paths in cmake fix

### LINKED ISSUE #116: *.option is searched in the execution directory instead of directory where the *.proto is
```
What steps will reproduce the problem?
1. create a make file project where the make file and the *.proto / *.option 
file is not in the same directory
e.g.
Build/Makefile
Config/alltypes.proto
Config/alltypes.option
2. command in the Makefile is:
$(NANOPB_SRC_GEN)/%.pb.c $(NANOPB_SRC_GEN)/%.pb.h: $(NANOPB_INPUT)/%.proto 
$(wildcard $(NANOPB_INPUT)/%.options)
    $(PROTOC) -I$(NANOPB_INPUT) --nanopb_out=$(NANOPB_SRC_GEN) $<
3. by calling the generator the *option file is ignored
4. change the command in the Makefile to:
$(NANOPB_SRC_GEN)/%.pb.c $(NANOPB_SRC_GEN)/%.pb.h: $(NANOPB_INPUT)/%.proto 
$(wildcard $(NANOPB_INPUT)/%.options)
    cd $(NANOPB_INPUT); $(PROTOC) -I$(NANOPB_INPUT) --nanopb_out=$(NANOPB_SRC_GEN) $<
the difference is the "cd $(NANOPB_INPUT); " before the generator call
5. the *.option file is used

What is the expected output? What do you see instead?
I expect as documented that the *option file is searched in the same directory 
as the *.proto file.

What version of the product are you using? On what operating system?
0.2.7 on Windows

Please provide any additional information below.


```

Original issue reported on code.google.com by `andreas....@gmail.com` on 30 Apr 2014 at 1:06

maciejmatuszak:
The cmake support checks *.options files and adds plugin include option "-I" so the plugin can find it.
Currently the path is relative and cmake will assume it is relative to CMAKE_CURRENT_BINARY_DIR. Since this is source file it should be relative to CMAKE_CURRENT_SOURCE_DIR. 
I guess it will work if you build in the source dir but not in general case.
This will fix #116 if you use cmake.

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -216,7 +216,7 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
     set(NANOPB_OPTIONS_DIRS)
 
     # If there an options file in the same working directory, set it as a dependency
-    set(NANOPB_OPTIONS_FILE ${FIL_DIR}/${FIL_WE}.options)
+    set(NANOPB_OPTIONS_FILE ${CMAKE_CURRENT_SOURCE_DIR}/${FIL_DIR}/${FIL_WE}.options)
     if(EXISTS ${NANOPB_OPTIONS_FILE})
         # Get directory as lookups for dependency options fail if an options
         # file is used. The options is still set as a dependency of the


PetteriAimonen:
Sounds reasonable.

maciejmatuszak:
@PetteriAimonen Thanks for quick turn!

PR #416, opened 2019-07-31
0.3.9.3 ftpointer size check

Gbillou:


PetteriAimonen:
Hmm, this looks to be against the maintenance branch while the pull request is against master branch. That causes the commit list to look a bit weird, but that is easy to fix.

However, the field structure has changed a lot in the master branch. And the maintenance branch is in bugfix-only mode. As is, there isn't a place where I can merge this. It might be an useful feature for the master branch, but I'm not sure if the need is common enough to warrant it. I think most people would add such checks by overriding `pb_realloc`, though that does not give such fine per-field control.

Adding new parameters to all macros is also quite invasive for just one feature.

PetteriAimonen:
I have decided that in current form I'm not going to merge this feature to nanopb main code base, as it requires too invasive changes.

The idea may be useful in some situations, but the implementation would have to be less invasive to keep the code size small & maintainable.

PR #418, opened 2019-08-06
Update readme

ElectricRCAircraftGuy:
Update readme with Google protobuf installation notes, as well as more clarity and info on compiling .proto files, including demonstrating both the 1-command `protoc` option as well as the 2-command option which calls a python script directly.


PetteriAimonen:
Thanks

Issue #426, opened 2019-08-27
add option to not encode scalar arrays as packed


prusnak:
Due to https://github.com/nanopb/nanopb/blob/95d614263cdb1cca3af26a9f2f7b02072e99eb0f/pb_encode.c#L121 it is not possible to send unpacked arrays of scalar datatypes.

There are still some codebases stuck on the old version of protobuf.js which does not recognize packed arrays (see https://github.com/protobufjs/protobuf.js/issues/432).

It would be nice to have an option to override this setting (which essentially changes the condition above to `if (0)`). I have tested this change in my codebase and this modified code correctly produces the unpacked scalar arrays which older protobuf.js has no trouble parsing.

PetteriAimonen:
Hmm, I guess it would be ok to have a compile-time `#ifdef` for that.

prusnak:
Sure, I am fine with an `#ifdef`. Will send a PR if you want.

PR #427, opened 2019-08-27
Introduce new compile time flag: PB_ENCODE_ARRAYS_UNPACKED

prusnak:
Don't encode scalar arrays as packed. This is only to be used when
the decoder on the receiving side cannot process packed scalar arrays.
Such example is older protobuf.js.

Fixes https://github.com/nanopb/nanopb/issues/426


PetteriAimonen:
Looks great, thanks!

prusnak:
Is this worth backporting to the 0.3 branch and releasing as 0.3.9.4?

PetteriAimonen:
Merged to 0.3 branch now, as it is safe enough change. But by itself, I don't think it merits a new release. I'll see if there might be enough reason otherwise for a new release some time in near future.

I guess it would also be time to finally get 0.4.0 out as well.

prusnak:
Okay, thanks for the update!

PR #438, opened 2019-10-10
Fix "unused variable" warning when using PB_ENCODE_ARRAYS_UNPACKED

prusnak:
In the last PR I forgot to add this change. Otherwise, the build fails with `-Werror` because `-Wunused-variable` is emitted when PB_ENCODE_ARRAYS_UNPACKED is used.

This should be backported into `maintenance_0.3` branch as well.

### DIFF
--- a/pb_encode.c
+++ b/pb_encode.c
@@ -124,7 +124,9 @@ static bool checkreturn encode_array(pb_ostream_t *stream, pb_field_iter_t *fiel
 {
     pb_size_t i;
     pb_size_t count;
+#ifndef PB_ENCODE_ARRAYS_UNPACKED
     size_t size;
+#endif
 
     count = *(pb_size_t*)field->pSize;
 


prusnak:
Related: Maybe we should also enable `-Werror`, `-Wall`, `-Wextra` in the CI? (That's what I use to build nanopb in my project but missed that when sending the last PR). It wouldn't catch this particular bug, because it is never triggered unless PB_ENCODE_ARRAYS_UNPACKED is defined, which is not at the moment.

PetteriAimonen:
Hmm yeah, it would be great if you can add a test case for this.

The `no_errmsg` test should be a good base, you only need to copy the SConscript and add the define there.

prusnak:
Added test in for PB_ENCODE_ARRAYS_UNPACKED in another commit

PetteriAimonen:
Great, thanks!

PetteriAimonen:
Merged to 0.3 maintenance also.

prusnak:
Great!

PR #447, opened 2019-10-30
install shared library(dll) in windows to the binary folder

der-imp:
When activating the cmake option `BUILD_SHARED_LIBS`  the created dynamic library ("protobuf-nanopbd.dll") in windows is not installed into the bin folder.


PetteriAimonen:
Sounds reasonable.

But note that using nanopb as a dll is kind-of unsupported, because currently the header files and code are quite tightly integrated. So after version upgrades you have to be careful to keep .dll version and the headers you compile with in sync.

PR #450, opened 2019-11-08
Allow comma-separated option parsing to handle "#include" in options

mark64:
When calling `protoc --plugin=protoc-gen-nanopb=protoc-gen-nanopb-plugin --nanopb_out=. '--nanopb_opt=--library-include-format=#include"%s"' --nanopb_opt=--no-strip-path --nanopb_opt=-xgenerator/nanopb/options.proto`, The following is given to `nanopb_generator.py` as `params`:

```--library-include-format=#include"%s"',--no-strip-path,--nanopb_opt=-xgenerator/nanopb/options.proto```

This passes the `len(args) == 1 and ',' in args[0]` check, so it then gets lexed. However, since there's a `#` character, `shlex` discards everything but: `--library-include-format=`.

Setting `lex.commenters = ''` allows `--library-include-format` to be specified in the above manner.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1855,6 +1855,7 @@ def main_plugin():
         lex = shlex.shlex(params)
         lex.whitespace_split = True
         lex.whitespace = ','
+        lex.commenters = ''
         args = list(lex)
 
     optparser.usage = "Usage: protoc --nanopb_out=[options][,more_options]:outdir file.proto"


mark64:
A few additional things:
1. If I add a space like `#include "%s"` instead of `#include"%s"`, then there is no way to get the parser to work correctly. The initial parse will split along the space, unless I add some escaped quotes. But if I do that, then the `shlex.shlex` step produces `['--library-include-format="#include "%s""', ...]` (with an extra wrapping quote around the include). I assume it's not okay to break compatibility with the `shlex.parse` format and comma-separated parsing is only offered as a secondary method, so I can live without space support.

2. How should I add tests for this?

PetteriAimonen:
Hmm, first time I hear about `--plugin-opt=` for protoc.

If you figure out a proper way to split those options in plugin mode, feel free to replace shlex with it.

Indeed currently there are no test cases for generator options.
To add them, one would:

1. Make a new directory, `tests/protoc_options` or something
2. Take e.g. `tests/options/SConscript` as a template. Add `opts = env.Clone()` and `opts.Append(PROTOCFLAGS = '''--nanopb_opt=--library-include-format=#include"%s" ''')`
3. Build some `.proto` that this affects, e.g. `opts.NanopbProto("testproto")`
4. Use the `opts.Match(...)` mechanism to check that correct patterns go into the .pb.h file. The `expected` file is a list of regexp patterns.

PetteriAimonen:
Thanks, merged. No reason really to postpone, it's better than old version anyway.

PR #451, opened 2019-11-09
Fix clang-tidy warnings about using signed integers in binary bitwise operations

torfinnberset:
This fixes clang-tidy warnings in the user's code when using nanopb encoding and decoding functions. The clang-tidy warning is related to https://wiki.sei.cmu.edu/confluence/display/c/INT13-C.+Use+bitwise+operators+only+on+unsigned+operands.

There are still many cases of this in the nanopb library, but this PR starts with just fixing the ones that end up visible in the user's code.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -164,58 +164,58 @@ typedef uint_least8_t pb_type_t;
 /**** Field data types ****/
 
 /* Numeric types */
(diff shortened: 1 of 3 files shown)


PetteriAimonen:
Thanks!

PR #454, opened 2019-11-13
Split nanopb_out command on FindNanopb.cmake

sdorre:
I have an issue when using NanoPB with Cmake in Windows. 

I want to specify an options file (Or just if the options file is in the same folder as the proto file).
The issue is that Windows is using semicolon (`:`) in path, and that interfere with the nanopb_out parsing. 

Example: Using cmake, the final command becomes :
`protoc.exe --plugin=protoc-gen-nanopb=C:/dev/build/nanopb/generator/protoc-gen-nanopb "--nanopb_out= -IC:/dev/build : C:/dev/build" C:/dev/src/data.proto`

As we can see. There is several `:` in the nanopb_out option, which lead to an error saying 
`unkonwn path /dev/build:C:/dev/build` as the parsing takes `-IC` as option and everything after is considered as the output buffer.

Solution:
NanoPB plugin options can be provided using "--nanopb_opt". 
Spliting the command here solve the issue.

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -256,7 +256,8 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
       ARGS -I${GENERATOR_PATH} -I${GENERATOR_CORE_DIR}
            -I${CMAKE_CURRENT_BINARY_DIR} ${_nanopb_include_path}
            --plugin=protoc-gen-nanopb=${NANOPB_GENERATOR_PLUGIN}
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks, looks reasonable.

The #450 pull request may still affect --nanopb_opt parsing in some situations, but for most cases it should work fine already.

PetteriAimonen:
Looks like this might need a version check for protoc, as it looks like at least protoc 3.6 doesn't support this option yet.

PR #458, opened 2019-11-28
change generator to raise error on unmatched options and add cli option to disable

### LINKED ISSUE #456: nanopb continues to compile if options have unmatched fields
`Following patterns in <filename>.options did not match any fields: <list of field names>`

If you specify options on a certain field name, and the generator cannot find it, it will print the above message, however the compilation process continues and files are still produced. I believe the correct behaviour should be for the process to terminate after that error message. If someone were to rename a field in their .proto file, and forget to do the equivalent rename in their .optiosn file, they would still get a compiled output, but with incorrect options.

wmmc88:
closes #456

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1681,6 +1681,10 @@ optparser.add_option("-f", "--options-file", dest="options_file", metavar="FILE"
 optparser.add_option("-I", "--options-path", dest="options_path", metavar="DIR",
     action="append", default = [],
     help="Search for .options files additionally in this path")
+optparser.add_option("--error-on-unmatched", dest="error_on_unmatched", action="store_true", default=True,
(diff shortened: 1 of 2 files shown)


wmmc88:
@PetteriAimonen  I fixed one of the tests, but im not sure why the name mangling tests fail. It should be able to pick up those fields, so i'm not sure why it doesn't.

PetteriAimonen:
Please make it off by default.

I value backwards compatibility a lot, and this has the potential to break otherwise working builds.

wmmc88:
updated the default behaviour to be backwards compatible

PetteriAimonen:
Great, thanks!

wmmc88:
Will this be part of 0.4.0 or after that?

PetteriAimonen:
It will go in 0.4.0, once I get around to releasing it.

Issue #401, opened 2019-05-23
pip-package for generator plugin


kaspar030:
I'm currently working on integrating nanopb as a package into [RIOT](https://github.com/RIOT-OS/RIOT/pull/11565).
While writing the corresponding README on how to get started with nanopb, I realized that getting the generator installed requires quite a lot more manual work.

Isn't the generator basically a python application? It could be made installable via "pip".

If there's interest, I can try PR'ing something.

PetteriAimonen:
Hmm, that's actually a great idea. I've never thought about it before, but it makes perfect sense.

I don't have much time to work on it in the near future, so if you can make a PR that would be great!

PR #441, opened 2019-10-18
Switch to using Python 3 for code generation, as Python 2.7 is EOL

torfinnberset:
Nanopb is currently still using Python2 via `nanopb_generator.py`. This PR changes the shebang to look for python3 in the environment, just in time for Python2 to reach end-of-life.

PR #460, opened 2019-12-12
Nanopb Generator as a python package

theunkn0wn1:
This Pull request allows nanopb's generator to be built into a python package.

#   What functionality it fixes/adds.
### additions
- Adds `poetry` as a development dependency, to build wheels from the `generator` folder.
- `pyproject.toml` project configuration file - describes the package as a whole and how to build it. (sdist)
- `nanopb_generator` script path, allows `generator/nanopb_generator` to be invoked from the CLI from any directory.
```
root@0a250617787d:~# ls
root@0a250617787d:~# nanopb_generator
Usage: nanopb_generator.py [options] file.pb ...

Options:
  -h, --help            show this help message and exit
  -x FILE               Exclude file from generated #include list.
# output truncated for brevity
```
- `protoc-gen-nanopb` script path, allows `protoc` to automatically detect the nanopb plugin. this means the plugin doesn't need to be explicitly declared!
```
root@0a250617787d:~# ls
empty.proto
root@0a250617787d:~# protoc --nanopb_out=. empty.proto
root@0a250617787d:~# ls
empty.pb.c  empty.pb.h  empty.proto
root@0a250617787d:~# cat empty.pb.c
/* Automatically generated nanopb constant definitions */
/* Generated by nanopb-0.4.0-dev */

#include "empty.pb.h"

/* @@protoc_insertion_point(includes) */
#if PB_PROTO_HEADER_VERSION != 40
#error Regenerate this file with the current version of nanopb generator.
#endif

PB_BIND(rover_empty_Empty, rover_empty_Empty, AUTO)



/* @@protoc_insertion_point(eof) */
```
## building the wheel
```
>  poetry build
Building nanopb (0.3.9.4)
 - Building sdist
 - Built nanopb-0.3.9.4.tar.gz

 - Building wheel
 - Built nanopb-0.3.9.4-py2.py3-none-any.whl
```
### fixes
`generator/nanopb_generator.py`
 - adjusted to use a relative import from its package, to make it work with python packaging.
#    How can the problem be reproduced / when would the feature be useful.
This PR allows nanopb, when installed as a python package. to be transparent to the end-user when generating sources.

I, for one, found automating NanoPB in my downstream build process to be somewhat frustrating, due to the requirement of embedding nanopb's git repository / binary distributions. Having an install-able python wheel allows downstream users to avoid this. 

This PR also allows nanopb's generators to target python3 as a side-effect, without effecting the shebang line (see #441 )

## Problems
For this pull request I took a minimalistic approach, only changing the bare-minimum to achieve the feature. 
Presently, the Nanopb generators get installed into the `generators` package, preferably I would like to place things under `nanopb.generators`. Unfortunately It would appear that we would need to move the `/generators/` folder to `/nanopb/generators` to achieve that result.

Oh, and please do review the `pyproject.toml` and make sure I got the author information correct.

---------

Closes: #401 
Relates to: #441

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -54,6 +54,9 @@ except TypeError:
          ****************************************************************************
     ''' + '\n')
     raise
+except ImportError:
+    # Invoked directly instead of via installed scripts.
+    import proto.nanopb_pb2 as nanopb_pb2
 except:
     sys.stderr.write('''
          ********************************************************************
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -54,8 +54,8 @@ except TypeError:
          ****************************************************************************
     ''' + '\n')
     raise
-except ImportError:
-    # Invoked directly instead of via installed scripts.
+except (ValueError, SystemError):
+    # Probably invoked directly instead of via installed scripts.
     import proto.nanopb_pb2 as nanopb_pb2
 except:
     sys.stderr.write('''
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -36,7 +36,7 @@ except:
     raise
 
 try:
-    import proto.nanopb_pb2 as nanopb_pb2
+    from .proto import nanopb_pb2
 except TypeError:
     sys.stderr.write('''
          ****************************************************************************
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,22 @@
+[tool.poetry]
+name = "nanopb"
+version = "0.3.9.4"
+description = "Nanopb is a small code-size Protocol Buffers implementation in ansi C. It is especially suitable for use in microcontrollers, but fits any memory restricted system."
+authors = ["Petteri Aimonen"]
+license = "Zlib"
+packages = [
+    {include = "generator"}
+]
+
+[tool.poetry.scripts]
+nanopb_generator = "generator.nanopb_generator:main_cli"
+protoc-gen-nanopb = "generator.nanopb_generator:main_plugin"
+[tool.poetry.dependencies]
+python = ">=2.7"
+protobuf = "^3.11"
+
+[tool.poetry.dev-dependencies]
+
+[build-system]
+requires = ["poetry>=0.12"]
+build-backend = "poetry.masonry.api"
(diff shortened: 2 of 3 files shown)


PetteriAimonen:
Looks great so far!

pyproject.toml looks mostly correct. As far as I know, protobuf-3.6 is currently new enough for all functionality, so the dependency can be made a bit less strict. Also updating the version number in this file has to be added to `tools/set_version.sh` so that it gets updated automatically.

I could merge this as is, but I think it might be good to figure out some solution for the package name. I wonder if it would be possible to have some kind of build rule that copies the generator directory under another name before building the package?

theunkn0wn1:
It would probably work if the build system did something to the effect of
```bash
cd path/to/project/root
mkdir nanopb
cp generator nanopb
poetry build
```
I will look into this and the `tools/set_version.sh` when I get home from work this evening.

theunkn0wn1:
Adjusted the `tools/set_version.sh` tool to cover the `pyproject.toml` file.

I have modified the pyproject.toml file to create the package `nanopb.generator`, this will require the build system to copy some files around, unless you want to change project structure. 

@PetteriAimonen Would you like me to add a built step to Travis to effect this and build a wheel?
```bash
# assuming executing from VCS root
mkdir nanopb
touch nanopb/__init__.py
cp generator nanopb
poetry build
```

PetteriAimonen:
This looks good now, thanks a lot!

Do you happen to know how publishing packages on `pip` works? I guess that would be a good place for easy installation.

theunkn0wn1:
You should just be able to `poetry publish`, see https://python-poetry.org/docs/cli/#publish

Here is the short of it:
If you have not yet you will need to make an account on PyPI, under which you can publish packages.
Pip pulls, by default, from PyPI.

To publish to PyPI, you need to create whats called a `wheel`, Poetry handles this for you (`poetry build`).
You upload this wheel to the correct endpoint and you now have a release! (`poetry publish` handles this for you).

For the long answer: [here is the documentation on PyPi](https://pypi.org/help/#publishing), and is a good read for understanding how packaging in python works at a lower level.

PetteriAimonen:
Ok, I think it works now: https://pypi.org/project/nanopb/

I reorganized files a bit, moving them to `extra/poetry` and added build shell script for the copying. Also made a jenkins job that will automatically publish packages for dev & stable branches.

theunkn0wn1:
Pulled down master and ran the build script, Looks like its working as intended with the python package. 👍 

Would it be possible to backport these changes for 0.3x so I can `pip install` in my downstream projects? (either that or release pre-release builds for the platformio integration?)

PetteriAimonen:
I think I did the backport already, should be on PyPI also :)

theunkn0wn1:
so you did, i missed that. Thanks! ❤️

theunkn0wn1:
the 3x backport appears to be broken. Il file an issue ticket.

Issue #462, opened 2019-12-15
Use Python version of protoc from grpc.tool.protoc


PetteriAimonen:
Currently using nanopb generator requires installing Python-based nanopb_generator, protobuf-python library and the C++ protoc. In #460 nanopb_generator was made installable using `pip install nanopb_generator` and it also brings in protobuf-python as a dependency. But protoc still needs to be installed separately.

There exists a Python version of the protoc generator and it is available in `grpcio-tools` package. Using this the C++ binary dependency could be dropped entirely.

It would make sense that `nanopb_generator` when invoked directly could take `.proto` files and transparently pass them through grpc.tool.protoc. This would simplify running the generator a lot.

Some more details in #461 comments.

theunkn0wn1:
Been looking into this on my downstream project, using the `grpc-tools` protoc is straight-forward.

Here is an excerpt of how im wrapping protoc to generate sources

```py
from grpc_tools import protoc
...
def arduino_build(ctx):
    # protoc.main expects a list of arguments
    protoc_args = [
        "protoc",
        f"-I={ctx.obj.include_path.absolute()}",  # include path seems to be necessary
        f"--nanopb_out={ctx.obj.output_path.absolute()}",  # output path
        f"empty.proto",  # files we want to build go here (in the downstream this is procedural, shortened for brevity.
    ]
    ...
    # Invoke protoc.
    protoc.main(protoc_args)  # invokes grpc.tools.protoc to generate sources
```

PR #463, opened 2019-12-15
use grpcio-tools protoc instead of system protoc

theunkn0wn1:
This PR adjusts the `generator/proto/__init__.py` file to use the `grpcio-tools` shipped protoc instead of depending on an externally-installed system protoc.

Note:: for `grpcio-tools` to install on windows with python 3.8, [a release candidate must be used](https://github.com/grpc/grpc/issues/20615)
This is the reason the update to `pyproject.toml` pins a pre-release version as the minimum.

Relates to #462


theunkn0wn1:
Appears the build is failing because its not installing the dependencies defined in `pyproject.toml`.

Would you like me to adjust the Travis build to install the new dependency (`grpcio-tools`)?

PetteriAimonen:
Could it be made to detect in runtime which is available and use that?

There are probably people out there who have `protoc` installed but cannot install `grpcio-tools` for one reason or other.

Bonus points if the auto detection is in some function so that I can reuse it when I eventually want to make `nanopb_generator.py` call `protoc`-equivalent for input files.

theunkn0wn1:
Hm, I can move `grpcio-tools` to be an "extras" requirement", so its an optional install.

Yes, it is possible to make checking which protoc is available a function.
Which do you want attempted first, system protoc or the grpc-tools?

PetteriAimonen:
I'd say `grpc-tools` would be preferred, because it seems more likely to be compatible with whatever version of `python-protobuf` is installed.

theunkn0wn1:
I added a new file `nanopb._utils` that contains three functions, their names should be self-explanatory. 
`nanopb._utils.invoke_protoc` will attempt to use `grpc_tools.protoc`, should it exist. 
If it does not, it will fall back to system-provided protoc. 

`grpcio-tools` is now marked as an "extras" dependency, see 
https://packaging.python.org/tutorials/installing-packages/#installing-setuptools-extras

```pip install nanopb[grpcio-tools]```

PetteriAimonen:
Awesome, thanks!

PR #465, opened 2019-12-16
Introduce new compile time flag: PB_VALIDATE_UTF8

prusnak:
Check whether incoming strings are valid UTF-8 sequences. Slows down the string processing slightly and slightly increases code size.

### DIFF
--- a/docs/reference.rst
+++ b/docs/reference.rst
@@ -61,6 +61,9 @@ PB_ENCODE_ARRAYS_UNPACKED      Don't encode scalar arrays as packed.
                                arrays. Such example is older protobuf.js.
 PB_CONVERT_DOUBLE_FLOAT        Convert doubles to floats for platforms that do
                                not support 64-bit doubles. Mainly AVR.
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Hmm, I wonder if perhaps a less fast but smaller code size algorithm might be desirable. After all nanopb trades speed for size elsewhere also. Particularly on AVR, the array will go in RAM.

But even as it is, I'm ok with merging this once you consider it finished. After all it is usable for most platforms and an optional feature.

prusnak:
I changed the implementation to the one which does not use look-up table. Hopefully this better matches the goals of nanopb. Also I marked the PR as "ready to review" (i.e. it's not in draft anymore).

PetteriAimonen:
Looks good, thanks! I'll eventually add a few unittests for failing utf-8 cases also, but I can do that myself.

prusnak:
Cool, I wanted to add failing cases for unit tests, but it was not obvious to me how to add them. It would great if you added them! Thanks!

PR #479, opened 2020-01-10
Fixed dead image link in documentation (png->svg)

bolind:


bolind:
I'm a little green to github, so this pull request might not be formatted ideally. The only commit that's interesting is 3eace1c.

PetteriAimonen:
Hmm yeah, I guess the time is ready for this.

Previously the .png image was generated by the Makefile rule, so that the online docs ( https://jpa.kapsi.fi/nanopb/docs/ ) displayed correctly in browsers without .svg support. But those browsers are now so old that there is no need to support them, and referencing the .svg directly makes the image appear correctly in .rst readers and github.

I merged the commit 3eace1c manually, thanks!

PR #486, opened 2020-02-03
Fix Windows 10 temp file handling

devjoa:
The temp file handling gives a Permission Denied on Windows 10, when generating the *.pb file, due to that the file is already open. This fix deals with that problem.


PetteriAimonen:
Merged, thanks! 

(The travis failure was just a timeout)

PR #496, opened 2020-02-26
Correct a typo in date on the changelog file

dtwlin:
Signed-off-by: David Lin <dtwlin@google.com>


PetteriAimonen:
Oh, and I was trying so hard not to make that very mistake this year! :D

PR #497, opened 2020-02-27
Added type_override option, and its handling to generator

zukaitis:
I propose to add an option, which would allow to override type of variable in generated C structs.

It is useful, if you have many different types of (enum) fields, but want them to be of the same type in C. This allows to use a single method to update field values, or store pointers to these fields in an array.

Of course, using this option requires extra carefulness, because mismatching an override type will result in all sorts of issues. Although, anyone, who has a need for such options, will likely be aware of the possible consequences

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -417,6 +417,9 @@ class Field:
         else:
             raise NotImplementedError(field_options.type)
 
+        if field_options.HasField("type_override"):
+            desc.type = field_options.type_override
+
         # Decide the C data type to use in the struct.
         if desc.type in datatypes:
             self.ctype, self.pbtype, self.enc_size, self.data_item_size = datatypes[desc.type]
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Seems ok, but would it be possible for you to add a test case for this? It could probably be part of the current "tests/options" test case.

zukaitis:
Sure, I'll try to look into it next week

zukaitis:
I've managed to add a check for generation of this option to the test. Please let me know, if anything else is needed

PetteriAimonen:
Looks good, thanks!

CZEMacLeod:
Do you have an estimate of when this change will be released as it is not in the [0.4.1](https://github.com/nanopb/nanopb/releases/tag/0.4.1) release?

PetteriAimonen:
@CZEMacLeod Hmm, I guess I could try to get 0.4.2 out within next month or so. There are a few issues on the tracker that I'd like to get fixed before the release.

PR #499, opened 2020-03-12
Fix adding the -I option to protoc exec

IoTPanic:
## Issue

The newest version of protoc requires the -I flag to be used otherwise you get the following output-

```
File does not reside within any path specified using --proto_path (or -I).  You must specify a --proto_path which encompasses this file.  Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think).
```

## Solution

Line 1938 modified to pass a `-I` and the protobuf directory `'-I'+os.path.dirname(os.path.realpath(filename))` testing working with a .proto file on OpenSUSE linux. **Please note I was unable to run the scons tests so I would recommend that is done before a merge**.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1935,7 +1935,7 @@ def main_cli():
         if filename.endswith(".proto"):
             with tempfile.TemporaryDirectory() as tmpdir:
                 tmpname = os.path.join(tmpdir, os.path.basename(filename) + ".pb")
-                invoke_protoc(["protoc"] + include_path + ['-o' + tmpname, filename])
+                invoke_protoc(["protoc"] + include_path + ['-o' + tmpname, '-I'+os.path.dirname(os.path.realpath(filename)),filename])
                 data = open(tmpname, 'rb').read()
         else:
             data = open(filename, 'rb').read()


PetteriAimonen:
Hmm, I'm a bit unsure about this. Protoc include paths seem to be used in weird ways sometimes and it could cause problems if people have files with same name in multiple directories.

It might be better to have `['.']` as the default value for `optparser.add_option("-I", ...` instead. That way if `-I` is specified to `nanopb_generator.py`, it would override the default include path. But then again, that does not work exactly the same as this pull request, because the `.proto` doesn't necessarily exist in current directory.

IoTPanic:
Well would it be a true statement that 99% of people that their include path would be the same as the directory of their proto file? I don't use protobufs extensively so I am unsure but I would think adding a default include path as the directory of the file with a command line option for passing another value into -I. Something needs to be done since as it stands in master nanopb does not work with the newest release of protoc. Is -I a new option?

IoTPanic:
Or, make it a required option in nanopd just as it is in protoc, either way. I would like to get back into python, if you told me what you like as the owner I would be more than happy to make a different change.

PetteriAimonen:
Yeah, I agree something needs to be done. But I'm not yet sure what, so I'll need some time to think about it.

I think specifying `-I` to `nanopb_generator.py` should be usable as a workaround in the meanwhile? It should also continue to work if this change or other fix is merged in.

PetteriAimonen:
Now that I look into it, I think this was already fixed in nanopb commit 766a5efcdd7c5 for the `protoc` version that comes from `pip install grpio-tools`. Now it appears the same change to how `protoc` behaves is in the C++ version, so I've now expanded the same logic to work for that too, in commit f8a530d

Could you check if this fixes the problem for you?

IoTPanic:
I'm not a fan of it defaulting -I to `.` as it appears to be doing `subprocess.CalledProcessError: Command '['protoc', '-o/tmp/tmp099alff3/messages.proto.pb', '/home/samuel/Downloads/messages.proto', '-I.', '-I/home/samuel/Documents/nanopb/generator/proto']' returned non-zero exit status 1.` but having the -I option now with the generator fixes my issue and I was able to get nanopb working on a new system with a fresh compile of protoc thank you very much.

PetteriAimonen:
I'm not a big fan of protoc's include model either, I'd like it to be simpler like how most C compilers work. But it is what it is, and I thing for best compatibility nanopb should follow it quite closely. Simple projects having one or a few `.proto` files will be easy to build anyway. For more complex setups that may have hundreds of `.proto` files, it gets unnecessarily difficult if every protobuf library has its own include model.

I'm not actually sure what was causing problems for you, as to me it seems like protoc (the C++ version) still includes current working directory by default: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/command_line_interface.cc#L1321

But having it done on nanopb side shouldn't cause any trouble, so it is safe enough change.

IoTPanic:
I'm not sure, I've compiled the most recent version of protoc twice and I cannot compile `.proto` files without the `-I` button.

PetteriAimonen:
Hopefully the fix committed earlier is ok for everybody, so I'm closing this off.

Issue #360, opened 2018-11-06
Add bazel rules for nanopb generation 


silvergasp:
**What should the feature do?**

_Option 1_
Use proto_lang_toolchain() to define a toolchain overriding the default protoc library.
- This feature should allow for nanopb to be used with `cc_proto_library` when imported using Bazel.
- The proto_lang_toolchain should override the default library using `--proto_toolchain_for_cc=` compile flag as detailed [here](https://docs.bazel.build/versions/master/command-line-reference.html).

Documentation for this feature can be found [here](https://docs.bazel.build/versions/master/be/protocol-buffer.html#proto_lang_toolchain).

_Option 2_
An alternative which may be more suitable is to make a new skylark rule `cc_nanopb_library`, that behaves identically to `cc_proto_library`, except using nanopb. This rule would use the nanopb generator explicitly, rather than requiring a command line flag everytime you want to compile a nanopb target, overriding the default protoc implementation. Further this would allow for `cc_proto_library` targets to coexist with the `cc_nanopb_targets`.

**In what situation would the feature be useful?**
This is useful in ensuring that the generated libraries in a bazel project are directly in sync with the proto files. i.e. protoc never needs to be manually run to keep the generated files in sync.

_Option 1_
This would be useful for using nanopb exclusively as the generator for a C++/C Project

_Option 2_
This would be useful when building for a heterogeneous system, i.e. a system with low resources (requires nanopb) and a system with enough resources to use standard c++ protobuf libraries.

I personally believe that Option 2 is a better alternative, as this will work well with the heterogeneous systems I work on. I am open to both alternatives.

**NOTE:** I have already done some work on this, but thought I'd open up a discussion. I am happy to help develop this further. Though there a few bugs I could use some help ironing out.

PetteriAimonen:
I have no experience in Bazel at all. But there are already some build files for Bazel: https://github.com/nanopb/nanopb/blob/master/BUILD

Is this different somehow?

silvergasp:
The BUILD file that you have mentioned if I am not mistaken is for building the runtime, or supporting libraries that the generated code will use. 

Note that the BUILD file uses a `cc_library` target named `nanopb`. A cc_proto_library differs from this, because rather than pointing to existing `*.cc` files it generates `*.cc` files that are dependant on protobuf files.

Perhaps this can be explained by the code below.
```python
# Uses the generated `*.cc/*.h` files from the dependant foo_cc_proto `cc_proto_library`
cc_library(
[...]

PR #500, opened 2020-03-17
Add bazel rules for nanopb generation

silvergasp:
This pull request addresses the use case for automatically generating nanopb files using bazel targets addressing #360. This follows a similar pattern to the native bazel cc_proto_library rules.

Summary:
- Added rules directory with corresponding bazel rules
- Added bazel_simple in the examples directory
- Added import method that will work with bazel where the current import paths would not
- Added detection logic in generator to determine if it is being called from bazel and if so, run the plugin not the cli

Please feel free to critique this pull request, I am happy to keep working on this until you are happy with it.


examples/bazel_simple/WORKSPACE
```diff
@@ -0,0 +1,53 @@
...
+    requirements = "@build_stack_rules_proto//python/requirements:protobuf.txt",
+)
+
+load("@protobuf_py_deps//:requirements.bzl", protobuf_pip_install = "pip_install")
+
+protobuf_pip_install()
+
+git_repository(
+    name = "com_github_nanopb_nanopb",
+    commit = "766ee1cebf6e1500dd9bd94fc217abb554ef55e7",
```
PetteriAimonen: What should be done with the git hash here? Would it be possible to use the already checked out version of nanopb by default, and have a commented example of referring to a particular git hash? That would make automated testing of the build rules a bit easier.

generator/nanopb_generator.py
```diff
@@ -59,9 +59,15 @@
...
 except (ValueError, SystemError, ImportError):
-    # Probably invoked directly instead of via installed scripts.
-    import proto.nanopb_pb2 as nanopb_pb2
-    from proto._utils import invoke_protoc
+    try:
+        # Probably invoked directly instead of via installed scripts.
+        import proto.nanopb_pb2 as nanopb_pb2
+        from proto._utils import invoke_protoc
+    except ImportError:
+        # Invoked using bazel
```
PetteriAimonen: This starts to feel like a hack on top of hack to handle the import paths in different situations. But I have no better alternative to offer, my understanding of Python import mechanism is a bit lacking.

generator/nanopb_generator.py
```diff
@@ -2049,7 +2055,7 @@ def main_plugin():
 
 if __name__ == '__main__':
     # Check if we are running as a plugin under protoc
-    if 'protoc-gen-' in sys.argv[0] or '--protoc-plugin' in sys.argv:
+    if 'protoc-gen-' in sys.argv[0] or '--protoc-plugin' in sys.argv or '_bazel_' in sys.argv[0]:
```
PetteriAimonen: Can you elaborate on how bazel is invoking `nanopb_generator.py` that it needs this special mechanism?

BUILD
```diff
@@ -1,21 +1,22 @@
+package(default_visibility = ["//visibility:public"])
```
PetteriAimonen: Would it be possible somehow to have the bazel build files all in one folder, `bazel_rules` or something? There are now be files littered across `rules` and the root of the project and it's a bit messy to know which build system each belongs to.

examples/bazel_simple/WORKSPACE
```diff
@@ -0,0 +1,53 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "com_google_protobuf",
+    commit = "09745575a923640154bcf307fba8aedff47f240a",
```
PetteriAimonen: It seems that protobuf-python is being installed from `pip`, and presumably the grpcio-tools version of protoc could be installed similarly. What is the git checkout of the protobuf repo being used for?

examples/bazel_simple/WORKSPACE
silvergasp: In general, if it is an example the commit should remain constant. If it is part of documenation, then it is usually idiomatic to have this either as a TODO (e.g.`commit = "TODO",`) or alternatively it should reference the latest release. The motivation behind making using the git commit's in bazel is so that your dependencies are deterministic regardless of when you fetch it. Example idiomatic documentation from [rules_python](https://github.com/bazelbuild/rules_python).

I would happily include an automated test target, however this would likely look a little different.
```py
local_repository(
    name = "com_github_nanopb_nanopb",
    path = "../..",
)
The difference here is is references a path on your current file system in contrast to a remote repository. So it wouldn't require getting the latest commit.

File: generator/nanopb_generator.py (hunk 2 above)
Comment (silvergasp) [reply generator/nanopb_generator.py]: This is absolutely a hack, and to be honest, this was the biggest headache in getting this to work. It is bothering me so it's certainly something that I would like to improve upon in general. I will have a bit more of a look around at other projects. I feel that it should be possible to find a common method of importing that works with pip, bazel and checked out repos.

File: examples/bazel_simple/WORKSPACE (hunk 1 above)
Comment (PetteriAimonen) [reply examples/bazel_simple/WORKSPACE]: Hmm, but when people check out the nanopb repository and build the example, wouldn't they expect to build it against the revision they've just checked out?

As for testing, I like to have automatic builds of the examples exactly like users would build them, as otherwise it is quite easy to break them without noticing. Typically that's just a matter of adding a command to a Jenkins job and checking if there are any build errors. But if the example is building always against some old revision, it gives no information on whether the build rules still work with the newest version.

File: generator/nanopb_generator.py (hunk 3 above)
Comment (silvergasp) [reply generator/nanopb_generator.py]: This is simply a result of me misunderstanding your documentation. I had not spotted the shell/batch files that add the --protoc-plugin command line flags. For some reason I thought you where directly passing the nanopb_generator.py to protoc as a plugin. I can work around this by adding specific shell scripts for bazel. Or alternatively I could potentially split nanopb_generator.py into three smaller python files, nanopb_plugin.py, nanopb_cli.py, nanopb_lib.py. This would remove the need for proxy scripts altogether. Totally up to you, it's really personal preference at this point.

Comment (silvergasp) [reply generator/nanopb_generator.py]: TLDR; it's not needed, I can do better.

Comment (PetteriAimonen) [reply generator/nanopb_generator.py]: Splitting nanopb_generator.py is something I've considered myself to improve maintainability, but better to do that separately eventually.

Using either the flag directly or the shell scripts should be fine. Because of Python's portability, just giving `--protoc-plugin` to nanopb_generator.py would avoid dependency on any particular shell on the build machine. It is also possible to just use `nanopb_generator.py myfile.proto`, in which case nanopb itself calls protoc behind the scenes.

File: BUILD (hunk 4 above)
Comment (silvergasp) [reply BUILD]: I can, there are trade-offs either way. The WORKSPACE file in the root is not strictly needed, so that could be removed altogether. The BUILD file in the root can also be removed, however it comes at a cost in ease of use of use when referencing this project remotely. In other words removing the BUILD file in the root the requires the build file to be specified in the users WORKSPACE file. e.g.
This:
```py
git_repository(
    name = "com_github_nanopb_nanopb",
    commit = "766ee1cebf6e1500dd9bd94fc217abb554ef55e7",
    remote = "https://github.com/silvergasp/nanopb.git",
)
```
Would become this:
```py
new_git_repository(
    name = "com_github_nanopb_nanopb",
    commit = "766ee1cebf6e1500dd9bd94fc217abb554ef55e7",
    remote = "https://github.com/silvergasp/nanopb.git",
    build_file_content = """
package(default_visibility = ["//visibility:public"])

licenses(["notice"])

exports_files(["LICENSE.txt"])

cc_library(
    name = "nanopb",
    srcs = [
        "pb_common.c",
        "pb_decode.c",
        "pb_encode.c",
    ],
    hdrs = [
        "pb.h",
        "pb_common.h",
        "pb_decode.h",
        "pb_encode.h",
    ],
    strip_include_prefix = ".",
)
"""
)
```
The other alternative is to merge everything in the /rules directory, into a single /rules.bzl file, and then /rules/BUILD can be merged with /BUILD.

PetteriAimonen: So if I understand correctly, the `/BUILD` is now for building the C files, while `rules/BUILD` is for running the generator?

`/BUILD` and `/WORKSPACE` have been in place for long enough that other projects probably depend on them, so it's probably not a good idea to move them. If the other rules can be merged into `/BUILD`, that sounds good to me. It would also be great if you can add a comment on top of each of the files, something like "This file is for the bazel build system, see examples/bazel_simple" and if there is any useful extra info about the purpose of a particular file.

examples/bazel_simple/WORKSPACE
silvergasp: This is more of a matter of personal preference. So for example using this you can make use of a specific version of protoc, and that selection will be entirely deterministic and hermetic. The pip rules are not considered hermetic. There is a bit of a divided group of people at bazel on how deterministic tool-chains should be (i.e. should you use system installations? or use a determinsitically built toolchain). 

You are right in that, this is not necessarily 'required', my personal preference is that all toolchains should be as deterministic as possible, at the expense of downloads/compile time. 

For context [here](https://github.com/grailbio/bazel-toolchain) is a completely deterministic c++ toolchain for bazel. Which fetches the entire toolchain and the sysroot, and matches them against checksums.

I could potentially make this optional?

PetteriAimonen: The protobuf building & installation is quite heavy, so it being a mandatory dependency might be quite annoying.

The grpcio-tools version of protoc is Python-based AFAIK, so even if one would go the git route instead of pip install, it might be faster and easier to get working.

I think bazel has some kind of mechanism of providing rules from the upper level project, so might there be some kind of rule for "get me working protoc", that the user could then override to be even empty if protoc is already available from `apt-get` or similar.

PetteriAimonen:
Hmm, have I forgotten about this pull request? Looking at the history, was it so that this is ready to merge (apart from the rename conflict that is easy to fix)?

If it was me forgetting it, sorry :)

nate-tri:
As an external interested party, I'm having trouble understanding which of the review comments are "things needing fixing" versus "coversations about motivations and possible alternative methods".  I'm happy to help this along if the original author is no longer available, but I could use some guidance here @PetteriAimonen.  Can you perhaps resolve any "discussion" type comments, or call out what actions you would like taken?

PetteriAimonen:
@nate-tri What would help the most is a review from someone who understands bazel (I have no experience with it at all), and test it out.

examples/bazel_simple/WORKSPACE
simmonmt: Chiming in from the make-it-hermetic camp, this is idiomatic for Bazel (and for how it's used in Bazel's parent org). As examples, when you use Bazel to build Go code it downloads the entire Go SDK (so Bazel has its own repeatable copy rather than using the system copy), and when you run unit tests it downloads what looks to be the entire JDK. Heavy-weight, in this case, is in the eye of the beholder. :-) 

If you're wondering whether the typical Bazel user would be surprised to see it downloading all of these repositories just to enable a nanopb_proto_library rule or to build the generator, no, they would not be surprised. It would be considered totally normal.

simmonmt:
It looks like his changes were trying to do two separate things:
1. Make the generator build with Bazel
2. Create rules (the .bzl files) that use the generator from step 1 to build proto.c/h files.

The second step depends on the first. The rules exposed by the second step would be referenced by other Bazel workspaces to pull in this project, the generators, and so on. 

Regarding your question about putting all the Bazel stuff in one place, the rules in step 2 (and their tests) can certainly be confined to a single directory, but the BUILD.bazel files needed for step 1 should be colocated with the source being built.

In order to keep this change easy to reason about, it might be good to break it into two pieces. The first piece teaches Bazel how to build the generator in a py_binary rule, and would extend the work that's already been done to teach Bazel how to build the runtime. That's a change that's good in its own right, regardless of whether the second piece -- rules in bzl files that allow other Bazel workspaces to nicely generate proto c/h files.

simmonmt:
FYI https://github.com/nanopb/nanopb/pull/633 extracts BUILD.bazel and WORKSPACE formatting changes of this PR into a separate PR, thus minimizing churn in subsequent PRs.

generator/nanopb_generator.py
jxramos: I'm not 100% on this but I get the sense that bazel has some gaps in python support for the various techniques python uses to resolve imports, especially those that don't always mirror the paths on the file system. This bazel issue I filed comes to mind here https://github.com/bazelbuild/bazel/issues/11414

silvergasp:
Sorry for leaving this one open for so long. It got put on the back burner for some time, though now I have renewed interest in it. As things have changed in the Bazel world so much since I originally created this PR I've completely rewritten it. As a bit of a summary of what I have here;

- Some macros to handle dependencies ( There are three separate files two *_deps.bzl and a nanopb_workspace.bzl). There is annoyingly no easy way around the three separate files until the [bazel-central-registry](https://github.com/bazelbuild/bazel-central-registry) is up and running, which should make all this a lot tidier. Though, I thought about making a 'bazel' directory for all the bazel plugins, what are your thoughts on that?
- Some docs to describe how to use the new rules.
- The plugins and BUILD.bazel files to make it all work.
- A small wrapper to call the generator as a plugin. This ensures Bazel will handle the python binaries nicely without extra hacks.
- A github action for the Bazel build.

deeglaze:
Excited to see the renewed energy in this PR, looking forward to it @silvergasp :)

PetteriAimonen:
Sorry for leaving you hanging so long also :)

I'm a bit unqualified to review this, but here is a list of things that currently look weird to me:

1. [extra/requirements_lock.txt](https://github.com/nanopb/nanopb/pull/500/files#diff-9c40c8fa94f1e73a466e619baea7a00e40f36a186bb80f75d9e60f046e47bf3c) and the long list of hashes there. Is it necessary to lock the versions? I have already test cases in place that verify that things work with newest grpcio-tools version. I do agree that there is a risk of breaking old builds if newest grpcio-tools changes something. But if that lock is just left there, no-one will be updating it and the bazel rules will pull outdated grpcio-tools.
2. [generator/bazel_generator.py](https://github.com/nanopb/nanopb/pull/500/files#diff-01dcc767f95031c29764c62e36a97fcd16dca41c1bad7dadf5e7fc0adbf78953) Is this wrapper necessary? Nowadays I feel like calling through `nanopb_generator.py foo.proto` is preferrable, and let it invoke protoc internally. But I guess this is taking advantage of some shared build rules for `.proto`? Could it just use `generator/protoc-gen-nanopb` instead of adding a new wrapper? I think the current bazel rules would also work without extra wrapper if the binary name generated by bazel was `protoc-gen-nanopb`, though I don't know what `py_binary` actually does.
3. Is it possible to move the `.bzl` files to subdirectory somehow? At the very minimum, they will need to have a header that identifies the purpose of the files and that they are specific to a particular build system which is not mandatory for nanopb.

silvergasp:
No worries;

## RE: 1
It's not technically necessary to lock the versions. But I would say it is idiomatic or even expected that Bazel deps are pinned/locked. For example if I use nanopb from bazel I would use something like this in my WORKSPACE file;
```python
git_repository(
    name = "nanopb",
    remote = "https://github.com/nanopb/nanopb",
    commit = "0894977c1c6a5048cadd00e1c0199dd701b42346",
)
```
Note here that I am using the full commit sha to pin nanopb at a specific point in the git history. So I would expect that when I fetch all the dependencies for the Bazel rules I fetch a version of gprcio-tools that works with that specific version of nanopb. Hence why it's useful to pin grpcio-tools to an older version.

Having said all of that you raise a valid point that these pinned deps could get out of sync, with mainline nanopb and cause a breakage. There is a tool included in this PR to update the lockfile to match the versions in the requirements.txt. However as you are not all that familiar with Bazel this might become an additional maintenance burden to occasionally run that tool.

How do you feel about me adding a github action to update the lock file and open a PR once a week/month?

## RE: 2
>  But I guess this is taking advantage of some shared build rules for .proto?

Yeah that's right, so you build a proto_library, target which creates the proto descriptor binary files. Which is then passed to protoc alongside a protoc plugin. It might be technically feasible to get this working using the nanopb_generator.py, but it would require fairly significant deviations from what would be the normal workflow for a Bazel/protoc rule. I think the way it works now is reasonably idiomatic.

> Could it just use generator/protoc-gen-nanopb instead of adding a new wrapper?

I think that it could, but this complicates things significantly in terms of the Bazel build. Each build action (e.g. compiling a c file, or a proto file) operates in a sandbox, so only the dependencies that you declare get added to the sandbox. It becomes rather complicated to get Bazel to create a sandbox for a shell script that then calls a python binary and have the paths workout to, for all the dependent python libs. It is something that I did try, though I kept running into edge cases and roadblocks where I couldn't get it working.

How do you feel about me rewriting `generator/protoc-gen-nanopb` to be a python script rather than a shell script? I think that would keep the Bazel build happy and should keep everything else working?

>  I think the current bazel rules would also work without extra wrapper if the binary name generated by bazel was protoc-gen-nanopb, though I don't know what py_binary actually does.

I hadn't thought about that, I'll explore this further.

## RE: 3
Yeah I can move the '.bzl' files. The only files that need to stay where they are, are `WORKSPACE` and `**/BUILD.bazel`.

PetteriAimonen:
1. Ok. I think I can just add the updating to my release process. Is `bazel run //extra:requirements.update` the command to use or something else?
2. Huh, I didn't realize that `protoc-gen-nanopb` was still a shell script. I'm totally fine with converting it to Python!
3. That would be great, it makes it easier to distinguish what files are relevant for what build system. `extra/bazel` would be a good directory name.

silvergasp:
> Ok. I think I can just add the updating to my release process. Is bazel run //extra:requirements.update the command to use or something else?

Yeah that's the command, that'll regenerate the lockfile. The only other thing that might need to be done is to change the version number in the requirements.txt otherwise regenerating the lockfile won't do much. Out of interest how are you handling dependency versions. Would it make sense to move that requirements.txt up a level?

If release notes works for you then that's great. Otherwise there is a [pull-request](https://github.com/marketplace/actions/create-pull-request#:~:text=A%20GitHub%20action%20to%20create,add%20files%20to%20your%20repository.) github action. Which it looks like I could get to trigger whenever the requirements.txt file is updated, rather than the once a month kind of thing.

>Huh, I didn't realize that protoc-gen-nanopb was still a shell script. I'm totally fine with converting it to Python!

Cool, I'll do that, If I'm not mistaken, the bazel_generator.py should be a drop-in replacement, so I'll just rename it.


I'll have those changes up shortly. I've just gotta fix up my environment a bit as grpcio_tools doesn't seem to like the latest python release.

silvergasp:
Ok, so it's looking like the tests are all passing again. Ready to have another go at a review if you like.

silvergasp:
Oh never mind, I missed something.

silvergasp:
So I think I've addressed all the points that you've mentioned and everything is good to go at this point.

This might come across as a little blunt. It seems that you don't want to merge this PR, should we consider closing this and exploring other avenues?

I should note that it's your project, you have to bear the maintenance burden, so there aren't any hard feelings if you just say "no, I'm not going to merge this". I am still very grateful for all the hard work that you've put into nanopb, and that's not going to change whatever the outcome is here. This PR has been open for more than 2yrs, a large part of this is because I've been rather slow to respond to reviews. In any case, it would be great if we can bring this to a conclusion.

Here's what I see as the barriers to getting this merged;
- There isn't enough expertise in the Nanopb team to review this change
- There is an extra maintenance burden associated with ongoing support for this change

# Alternative 1
I can limit the surface area of this change down to;
- The python rewrite of the plugin protoc-gen-nanopb
- The changes in the root level build file.

I think I could do this by modifying 3 files with no extra additions.

This would involve removing all the vendors and the *.bzl files. This would allow me to create a separate repository, with the bazel plugins. e.g. this would probably be under the [bazelembedded](https://github.com/bazelembedded/bazel-embedded) organisation, and I'd call it something like 'rules_nanopb'. This would result in a fairly minimal increase to the 'Bazel' surface area in this repository and I can take on the maintenance burden elsewhere.

# Alternative 2
I maintain a fork of this repository, with the Bazel rules. This is what I'm currently doing, but it's rather cumbersome, and I'd like to avoid continuing this if possible.

PetteriAimonen:
Sorry, I still *do* want to merge this.

I seem to have misinterpreted your previous comment "I missed 
something." I thought there was still something you wanted to change.

I've been on vacation lately and trying to avoid computers, but I'll try 
to find time to review this in near future.

silvergasp:
Oh yeah, my bad. I can see how that could be misinterpreted. Cool well, there isn't any serious hurry, so enjoy your vacation. This PR will still be here when you're ready to look over it.

PetteriAimonen:
Thanks, indeed everything was perfectly ready for merge!

I also got bazel installed on my own machine and the requirements update command working without problems.

silvergasp:
Awesome! Feel free to tag me in any Bazel-related issues/PRs, and I'll do my best to help out with reviews or bugs.

PetteriAimonen:
Thanks! Yeah, I usually try to tag the latest people who have touched particular build rules.

gbalke:
My mind is blown that this got merged the same day I was looking up if bazel was supported :laughing:

Thank you @silvergasp @PetteriAimonen :pray:

PR #504, opened 2020-03-26
Regression test + fix for proto3 submessage improperly considered empty

Gbillou:

See the attached test case

**What happens?**

On `pb_encode`, `pb_check_proto3_default_value` may improperly consider submessages as being empty (set to  default value) when FT_POINTER fields are used.

It is only checking the `field->data_size` first byte(s) of the pointer, which for strings or bytes is 1, so if an pointer happens to be aligned on 0x100 it'll be considered as being NULL.

**What should happen?**

Non empty submessages should always be serialized.

**Notes**
Fixing PB_SINGULAR_POINTER `field->data_size` to actually be a full pointer size seems to work without breaking other tests cases `pb_membersize(st, m[0])` -> `pb_membersize(st, m)`. But this leads to OOBW on `pb_decode`.

This bug is not present in 0.4 and onward thanks to `has_submessage`.

### DIFF
--- a/pb_encode.c
+++ b/pb_encode.c
@@ -304,6 +304,12 @@ static bool pb_check_proto3_default_value(const pb_field_t *field, const void *p
             return true;
         }
     }
+
+    /* Compares pointers to NULL in case of FT_POINTER */
+    if (PB_ATYPE(type) == PB_ATYPE_POINTER && PB_LTYPE(type) > PB_LTYPE_LAST_PACKABLE)
+    {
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Thanks! Fix merged in maintenance_0.3. Regression test merged to master branch also.

I think there may be actually some related bug in master branch also even though the test passes, but I'll have to review more carefully. The data_size is indeed not correct for comparing pointers byte-per-byte. I added that as issue #505.

Issue #506, opened 2020-03-27
find_package broken with cmake 3.17


twam:
We're currently using cmake's find_package to include nanopb in our project. Currently this is done using `find_package(Nanopb REQUIRED)`.
Starting with cmake 3.17 this throws the following warning:
```
CMake Warning (dev) at /usr/local/Cellar/cmake/3.17.0_1/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (NANOPB)
  does not match the name of the calling package (Nanopb).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
Call Stack (most recent call first):
  external/nanopb/extra/FindNanopb.cmake:341 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  xxx:27 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

In FindNanopb.cmake the find_package_handle_standard_args is currently called 
```
FIND_PACKAGE_HANDLE_STANDARD_ARGS(NANOPB DEFAULT_MSG
...
```
with NANOPB in uppercase. However if I change my include to
`find_package(NANOPB REQUIRED)`
it cannot find the Find*.cmake file anymore:
```
CMake Error at xxx.cmake:27 (find_package):
  By not providing "FindNANOPB.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "NANOPB", but
  CMake did not find one.

  Could not find a package configuration file provided by "NANOPB" with any
  of the following names:

    NANOPBConfig.cmake
    nanopb-config.cmake

  Add the installation prefix of "NANOPB" to CMAKE_PREFIX_PATH or set
  "NANOPB_DIR" to a directory containing one of the above files.  If "NANOPB"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  CMakeLists.txt:27 (include)
  ```

So I guess either the file needs to be renamed to FindNANOPB.cmake or we must use Nanopb when calling find_package_handle_standard_args.

PetteriAimonen:
Hmm, this is beyond my cmake understanding.

twam:
You should decide for one spelling of nanopb: `nanopb`, `Nanopb`, `NANOPB` and then we use it all the times :)

PetteriAimonen:
Ok :)

[...]

PR #507, opened 2020-03-27
Use Nanopb as cmake package name.

twam:
* As cmake find file is named 'FindNanopb.cmake' users are required
  to use `find_package(Nanopb)` to find nanopb. cmake >= 3.17 warns
  if package name (NANOPB up to now) and name used within find_package
  to not match. This changes to name to Nanopb. As the found variable
  is always with with UPPERCASE as well, this should be backwards
  compatible for users.
* Fixes Issue #506


PetteriAimonen:
Thanks!

PR #509, opened 2020-03-30
Use python3

twam:
* Switch to Python 3 as Python 2 has ended its support on 2020-01-01.
  (https://www.python.org/doc/sunset-python-2/)

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # kate: replace-tabs on; indent-width 4;
 
 from __future__ import unicode_literals
--- a/generator/protoc
+++ b/generator/protoc
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import sys
 from nanopb_generator import invoke_protoc


PetteriAimonen:
Duplicate of #441.

The problem is that I still want to run the tests against python 2 also. Currently this works quite well by just changing /usr/bin/python between python 2 and python 3 in the docker file. I guess I could symlink python3 to python2 in the docker, but that seems a bit silly. And anyway, completely dropping python 2 support would be a breaking change and as such post-poned until next major release.

But.. shouldn't distributions just switch /usr/bin/python to be Python 3?

twam:
I'm using nanopb on OS X and home brew removed their support for python2, so I'm kinda forced to use python3. Of course I can point /usr/bin/python to python3, but this would break other incompatible scripts. As those scripts work with python3 I think changing the shebang is the correct approach.

When are you aiming for the next major release?

PetteriAimonen:
When I run out of numbers for 0.4.x ;)  Probably still a few years away.

If you don't have Python 2 on the system, I don't see how the symlink would break anything worse than they are already broken. Or is the issue that you don't have the protobuf libs for Python 2? That's a bit annoying indeed. `pip3 install nanopb` would be one way to go about it, or `python3 nanopb_generator.py`. I didn't really think about the the python-protobuf dependency when writing the comment above, but that makes it even more difficult to switch the shebang - if someone has the protobuf library installed for Python 2, switching it now would suddenly break their builds without any easy to way to fix it.

[PEP 394](https://www.python.org/dev/peps/pep-0394/) is quite a mess these days.

twam:
OS X provides only python2.7 by default and sets python to this. However there is no pip provided by OS X, so installing python-protbuf is a pain. Almost everybody on OS X uses https://brew.sh to install tools like python and there only python3 with pip3 is provided.

At some point in time this will break for users and if the next major release is years in the future I suggest to do it rather now than later as currently the dying of python2 is in people's mind anyhow. But this is just my 2 cents.

PetteriAimonen:
Yeah, I don't know either what is best - all I know is that the change would break my test systems, and that is a quite good indication that it would break things for many users also. So far I've prioritized backwards compatibility quite strongly, but it is a good point that it is difficult to work around the problem on Mac OS X.

One way would be to have some kind of automatic fallback to Python 3 if python-protobuf cannot be imported, but I'm not sure if there is any sane way to accomplish that.

PetteriAimonen:
After consideration, I've made this change to newest git master and it will be part of 0.4.2 release.

PR #515, opened 2020-04-08
use split --nanopb_opt only on Windows

marekzukal:
to allow other systems without the ':' in their path to use older versions of protoc

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -249,6 +249,13 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
         set(NANOPB_PLUGIN_OPTIONS "${NANOPB_PLUGIN_OPTIONS} ${NANOPB_OPTIONS}")
     endif()
 
+    # There is a problem with ':' as option separator in protoc args so deal with it by splitting the options
+    # However this requires a version of protoc >= 3.6 so keep the old behavior for non-windows systems
+    if(CMAKE_HOST_SYSTEM MATCHES "Windows")
+        set(NANOPB_OPT_STRING "--nanopb_opt=${NANOPB_PLUGIN_OPTIONS}" "--nanopb_out=${CMAKE_CURRENT_BINARY_DIR}")
+    else()
+      set(NANOPB_OPT_STRING "--nanopb_out=${NANOPB_PLUGIN_OPTIONS}:${CMAKE_CURRENT_BINARY_DIR}")
+    endif()
     add_custom_command(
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Hmm, differentiating by platform doesn't seem like a good idea. After all, the same problem with options escaping and `:` exists also on Linux.

Would it possible to use some option or version detection instead to enable the old behavior?

marekzukal:
Well, I admit it is not a very clean solution, but it would solve most of the problems. I believe that the separator character is also kind of a "hacky" solution. I do not have any experience working with protoc. Does it pass the -I parameters to the plugin as well? Is there a possible solution without both ':' and --nanopb_opt?

Otherwise then, it should be possible to test the version of protoc, test for the ':' in the path and then pick a way to pass the opts the new way, old way, or fail with some readable message. Would that be acceptable? Do you know which version of the protoc is the first one to support it? I went through the changelogs and could not find it. There are different protoc versions mentioned around this project.

PetteriAimonen:
No, protoc doesn't pass `-I` options to plugins automatically, which is very unfortunate. I once tried to get that changed, but failed to manage to make a patch that would be accepted upstream. And indeed, the `:` protoc originally used to separate options was quite hacky, the separate `--plugin_opt` is better.

I've tested and know that the option exists on protoc 3.6. Looking at the upstream protoc github, it was added in this pull request in October 2016: https://github.com/protocolbuffers/protobuf/pull/2284 . Based on that, it should be in protoc 3.2 and newer.

marekzukal:
OK, I am detecting a protoc version and then choosing the way of passing opt or failing with a message explaining what is happening. 

However, I had to merge conflicting patch 9cd50b09b619a24977fed6deed4f43cd3ee4c5b7 and it broke things for me. cmake_relpath was working before, now I am getting files in sub/sub directory. Maybe we should not pass all the directories to protoc but only the ones for the currently processed file.

PetteriAimonen:
Ok, thanks for the info, I'll revert #516 for now.

I'm a total cmake novice and don't use it myself in any project, so reviewing how things should work in particular cases is a bit tricky.

marekzukal:
@PetteriAimonen , I have kicked out the change from #516 it should be OK now. I mean the paths work the same way they did before #516 . That should be dealt with still, if some users are experiencing problems. There is one line that needs to be changed and it is marked in the source by a comment. 

The commit history in this branch looks terrible. Do you need me to create a new branch with just one commit or are you able to merge it squashed?

PetteriAimonen:
Thanks, squashed and merged!

PR #523, opened 2020-04-19
Added the ability to supply an array and length rather than callbacks…

nimo-labs:
First of all, I appologise for VS Code doing a number on the formatting.
The key changes can be found as follows:
pb.h: Additional fields added to pb_callback_s
pb_encode.c: Additional code added to encode_callback_field()
pb_decode.c: Additional code added to decode_callback_field()

Callbacks are great for data heavy systems / transfers and a wonderful feature to allow an easy way for the library user to handle different I/O locations.

Strings are great for simpler messages, however as the name suggest, they handle strings. Using them for binary data falls over as soon as \0 is encounted in the data stream and requires some other way of passing the data length given that callbacks only accept one "arg" arg.

Adding the option of simply passing a length and pointer allows the programmer to create much neater code (at least in certain circumstances) as shown below.

For example this:
bool writeArray(pb_ostream_t *stream, const pb_field_iter_t *field, void *const *arg)
{
    if (!pb_encode_tag_for_field(stream, field))
        return false;

    return pb_encode_string(stream, (uint8_t *)*arg, strlen(*arg));
}

void protbufEncodeSend(nimocontrolUc_CommandType command, unsigned char *data, unsigned char dataLen)
{
    unsigned char pbArray[64];

    nimocontrolUc message = nimocontrolUc_init_zero;
    /* Create a stream that will write to our buffer. */
    pb_ostream_t oStream = pb_ostream_from_buffer(pbArray + 1, sizeof(pbArray) - 1);

    /* Fill in and encode the header data */
    message.command = command;

    /*Encode data if there is any*/
    if (NULL != data)
    {
        if (0x00 != *(data + dataLen))
        {
            printf("Error: ProtobufEncodeSend, data NOT null terminated!\r\n");
            return;
        }
        message.data.arg = data; /*This MUST be null terminated!*/
        message.data.funcs.encode = writeArray;
    }
    /* Now we are ready to encode the message! */
    if (!pb_encode(&oStream, nimocontrolUc_fields, &message))
    {
        printf("Encoding failed: %s\r\n", PB_GET_ERROR(&oStream));
        return;
    }
    usb_send(USB_EP_SEND, pbArray, sizeof(pbArray));
}

Becomes this:
        SimpleMessage message = SimpleMessage_init_zero;

        /* Create a stream that will write to our buffer. */
        pb_ostream_t stream = pb_ostream_from_buffer(buffer, sizeof(buffer));

        /* Fill in the lucky number */
        message.lucky_number = 13;

        message.dataField.usingPtr = 1;
        message.dataField.buffLen = 4;
        message.dataField.buffer = (unsigned char *)"1234";

        /* Now we are ready to encode the message! */
        status = pb_encode(&stream, SimpleMessage_fields, &message);


PetteriAimonen:
Well, the formatting mess-up makes this really hard to even review, so it is not going to get merged as-is.

For encoding, this is somewhat similar to what field type `FT_POINTER` already does. But I agree that there are small differences in how array length are stored, and in allocation in `FT_POINTER` mode. I have been considering adding another argument to `pb_callback_t` to simplify things, and the `callback_datatype` method actually allows even more customization but is probably a bit too complex for the general case.

The expansion of `pb_callback_t` is a bit undesirable also. Overall, I feel that this adds code and features, but may not be something that I want as part of nanopb core.

nimo-labs:
Sorry for the incredibly slow reply, my day job is keeping me very busy at the moment.

I agree with the formatting mess, do you have a particular coding style in use? I should be able to configure astyle to output any common style which should help with merges etc.

I'll have a look at the `FT_POINTER` type, I wasn't even aware that it exists to be honest, so it may do everything I need.

I have to be honest, the callback type is VERY complex at the moment, trying to get it working initially took at least as long as making the modifications to add the extra parameters to the callback type. Maybe it's a documentation thing? A simple example of callbacks without the extras of RPC etc. would help alot.

Assuming that `FT_POINTER` doesn't solve my use case, is there another way of implementing this feature that you would be happy with? I'm happy to do the actual implementation as I have here.

A compromise maybe to add a second `arg` to `pb_callback_t` so that the user could use to supply the length of the data in `arg`, however I think that would result in an interface that is almost as large as this approach and not as flexible.

nimo-labs:
I've spent some time today looking at `FT_POINTER` and can't even get nanopb to generate that type.
I've tried adding `(nanopb).bytes = FT_POINTER` to .options I've also tried `(nanopb).data = FT_POINTER` (data is the name of my field) and `bytes:FT_POINTER`, either way the data field is still generated as a callback. I know that nanopb is processing my .options file as if I add some rubbish to it then it generates warnings.

Frankly though I'm not convinced that `FT_POINTER` is a good idea anyway as https://groups.google.com/forum/#!searchin/nanopb/ft_pointer%7Csort:date/nanopb/TTolnt6DgMc/mqSt-8zsBQAJ suggests that you then need to enable malloc to use the resulting array, which makes sense however is still over kill vs being able to pass a pointer to a fixed block of memory, especially as you may be reusing that block for other uses on a memory constrained UC.

PetteriAimonen:
There are some good ideas here, but as it is the implementation does not feel like a good fit for nanopb and would become a maintenance burden.

PR #525, opened 2020-04-24
fix failed tests on Mac:

XuJiandong:
1. no -Wlogical-op available on Mac
2. use pointer (%p)


PetteriAimonen:
Hmm, I'd like to keep -Wlogical-op on GCC, as it is an useful warning.

Mac OS X for some reason aliases `clang` as `gcc`. The `scons CC=clang CXX=clang` shown in readme should work, but it might be nice to autodetect that also.

The pointer cast seems reasonable.

XuJiandong:
Confirmed that "scons CC=clang CXX=clang++" is working under Mac OS X.
It's not needed to merge this into master.

PR #534, opened 2020-05-19
Relax cmake python interpreter requirements to support Python3.

srfilipek:
This removes the Python 2.7 interpreter requirement for `cmake` and fixes issues when build systems may be using nanopb but no longer support Python 2.

* Run `cmake .` on a system without Python 2.7, or
* Integrate into a build system that installs required packages for Python3

**What happens?**
* cmake fails to find an appropriate python version, or
* Building fails since required packages may not be installed for Python 2.7

**What should happen?**
* cmake should use the default Python interpreter

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,9 +41,9 @@ endif()
 if(nanopb_BUILD_GENERATOR)
     set(generator_protos nanopb)
 
-    find_package(PythonInterp 2.7 REQUIRED)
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks, this seems reasonable. I still have to figure out whether to use Python 3 by default everywhere, but relaxing the version requirement in CMake definitely makes sense.

PR #537, opened 2020-05-28
Rename BUILD as BUILD.bazel

kalbasit:
Bazel [natively supports build files named BUILD and BUILD.bazel](https://docs.bazel.build/versions/master/build-ref.html#packages).
Using BUILD conflicts with CMake when building on case-insensitive file systems such as the default filesystem of macOS machines.


PetteriAimonen:
Great, I have actually wondered about the naming myself but didn't know it could be called `BUILD.bazel`.

Issue #530, opened 2020-05-06
Default values of +-Infinity and NaN not supported


jrullman:
A floating-point field marked with [default = inf] will result in generated code with default value 'inf', which doesn't compile. According to https://developers.google.com/protocol-buffers/docs/reference/proto2-spec#floating-point-literals, 'inf' and 'nan' are allowed values. The generator script should perform the following replacements on default values for floating point fields:

inf -> INFINITY
-inf -> -INFINITY
nan -> NAN
-nan -> -NAN

INFINITY and NAN are defined in math.h.

jrullman:
(INFINITY has to be cast to float in the single-precision case)

PetteriAimonen:
Sounds reasonable.

Currently `math.h` isn't being included, so the include should be added by the generator only if these default values are found in the file.

PR #538, opened 2020-05-28
Add support for infinity and nan floating-point defaults (#530)

jrullman:
Fix for https://github.com/nanopb/nanopb/issues/530.

The proto2 spec allows floating-point fields to have default values of 'inf', '-inf', 'nan' or '-nan'. The generator script copies these directly into source code, where they fail to compile (they do parse correctly in Python).

This change adds the following conversions for generated C code:
inf -> INFINITY
nan -> NAN
and adds an include for <math.h> where one of these occurs.

Note: The macros INFINITY and NAN are C99 features, so this won't work for strict ANSI C. However, the user can supply their own macro definitions in that case.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -354,6 +354,7 @@ class Field:
         self.ctype = None
         self.fixed_count = False
         self.callback_datatype = field_options.callback_datatype
+        self.math_include_required = False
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Looks great, thanks!

Issue #532, opened 2020-05-13
Decoding doesn't set default enum values correctly in proto2


jrullman:
According to https://developers.google.com/protocol-buffers/docs/proto#optional:
"For enums, the default value is the first value listed in the enum's type definition."

For a proto2 message containing an enum field, the "_init_default" struct initializer follows the above rule, but the decoder does not. No extern const "_default" value for the enum is generated, and the decoder will set the enum field to zero when the bytes don't contain that field.

PetteriAimonen:
Ah, this was missed when #295 was fixed.

A possible workaround is specifying the default value explicitly.

jrullman:
The most efficient fix would be to have a global for the default value of any enum (when it's not 0), rather than having to add a global for any field of enum type.

PetteriAimonen:
The way 0.4.x handles default values it would just take 1-2 bytes in the protobuf encoded representation of the defaults.

PR #539, opened 2020-06-01
Set proto2 enum fields to first value when no default is given, even if nonzero (#532)

jrullman:
Fix for https://github.com/nanopb/nanopb/issues/532.

According to https://developers.google.com/protocol-buffers/docs/proto#optional, when a proto2 enum field does not specify a default value, the default value is the first declared enum value, which may not be zero. For an enum with non-zero first value and a field with no default, the decoder was setting the field to zero, instead of the first enum value.


PetteriAimonen:
Thanks!

PR #549, opened 2020-06-22
Swift package manager

paulb777:
This is the Swift Package Manager configuration for nanopb we're using for initial Firebase testing with Swift Package Manager.

We may need to use nanopb from a fork because of our [versioning requirements](https://github.com/google/nanopb-podspec), but I'm happy to contribute to making this mergeable if there's interest.

### DIFF
--- a/.github/workflows/spm.yml
+++ b/.github/workflows/spm.yml
@@ -1,7 +1,7 @@
 name: spm
 
 on:
-  [push, pull_request]
+  push:
     paths:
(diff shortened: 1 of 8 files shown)


PetteriAimonen:
Thanks. Regarding versioning, I've been thinking of transitioning to semantic versioning, but not sure when. Perhaps on 0.5.0 :) Maintaining binary ABI compatibility is not currently in plans though.

PR #554, opened 2020-07-01
Minor style improvements to nanopb_generator.py

cxong:
- use in instead of multiple equals comparison
- remove unnecessary elif skipped by return


PetteriAimonen:
Thanks!

PR #555, opened 2020-07-02
Fix so python files are installed relative to CMAKE_INSTALL_PREFIX.

dflogeras:
This issue was addressed in bug #296, and recently broken again.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,7 +60,7 @@ if(nanopb_BUILD_GENERATOR)
         add_custom_target("generate_${generator_proto_py_file}" ALL DEPENDS ${generator_proto_py_file})
         install(
             FILES ${PROJECT_BINARY_DIR}/${generator_proto_py_file}
-            DESTINATION ${Python_SITELIB}
+            DESTINATION ${PYTHON_INSTDIR}
         )
     endforeach()
 endif()


PetteriAimonen:
@srfilipek Hmm, this was changed to `Python_SITELIB` in your commit 84941ce29644f33. Any ideas?

Can someone clarify to me the difference between `Python_SITELIB` and `PYTHON_INSTDIR`?

dflogeras:
As best I understand, here goes:

Python_SITELIB is a variable provided by CMake which points to the system-wide place where Python expects to find 3rd party modules.  For instance on my machine which has Python installed in /usr/, that is /usr/lib/python3.7/site-packages/.

PYTHON_INSTDIR is the same but without the prefix (as calculated in our CMakeLists.txt).  ie. for my system that would be just lib/python3.7/site-packages.  We then were using that to install relative to CMAKE_INSTALL_PREFIX. So if I choose to install nanopb into /home/me/nanopb/, I'd end up with something like

/home/me/nanopb/include/pb_encode.h (and others)
/home/me/nanopb/lib/libnanopb.a
/home/me/nanopb/lib/python3.7/site-packages/[generator related files]

If instead we install to Python_SITELIB, and the user does not have permission to write to /usr/ (for instance on a machine your office administers), you cannot install nanopb as a regular user.

If the user does have permission, and wishes to install system-wide, setting CMAKE_INSTALL_PREFIX=/usr/ will still achieve the same result as if we had used Python_SITELIB.

PetteriAimonen:
Ok, that makes sense. I'll see if @srfilipek has any thoughts about this, otherwise I'll just accept this and add a test case of cmake installation with non-root account.

dflogeras:
Alternatively, Python also can be queried for where the per-user site-packages is located (for me, /home/me/.local/lib/python3.7/site-packages, would be different on other platforms).

We could potentially offer a CMake option to install either there, or system-wide.  This would have the advantage of the user does not have to mess with PATH variables for python to find the generator .py files.

srfilipek:
Ah, I never caught this since I operate under a virtual environment. I can't recall or reproduce the situation that prompted the change in the first place for that particular line.

This patch looks fine and works in my testing.

PetteriAimonen:
Thanks!

PR #556, opened 2020-07-03
Install missing _utils.py and binary/wrappers.

dflogeras:
This patch installs the missing _utils.py and nanopb_pb2.py into the
expected proto/ subfolder.  Otherwise the generator script imports will
not find them in the proper namespace.

It also installs the generator and wrapper scripts in
CMAKE_PREFIX_PATH/bin/. WIN32 is handled specially since it has a
separate wrapper (.bat).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,9 +60,27 @@ if(nanopb_BUILD_GENERATOR)
         add_custom_target("generate_${generator_proto_py_file}" ALL DEPENDS ${generator_proto_py_file})
         install(
             FILES ${PROJECT_BINARY_DIR}/${generator_proto_py_file}
-            DESTINATION ${PYTHON_INSTDIR}
+                  ${generator_proto_file}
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Hmm, I think `nanopb.proto` might be needed also, so that `.proto` files that import it will compile without extra settings.

If you can, try e.g. `..../nanopb_generator.py tests/common/person.proto` to see if that works.

dflogeras:
OK, I just corrected this.  Please review, as I did it 'generically' by adding the rule to install the .proto file from within the foreach() although I'm not sure of the design.

Thanks for pointing me at this feature.  I've historically used .options, but have at least once gotten into trouble when someone regenerated the messages forgetting the .options files and silently broke everything.  Having the nanopb specific stuff directly in the .proto is a better solution IMHO.

PetteriAimonen:
Thanks. I'll try to add some kind of build test later if I remember.

PR #561, opened 2020-07-20
Adds support for grpcio-tool to FindNanopb.cmake

### LINKED ISSUE #560: Cmake generation not working with grpc-tools
I am trying to setup nanopb using cmake, but it fails in the generation phase requiring protoc.
```bash
CMake Error at /home/alex/.cmake-3.16.5-Linux-x86_64/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
  Could NOT find Nanopb (missing: PROTOBUF_PROTOC_EXECUTABLE)
Call Stack (most recent call first):
  /home/alex/.cmake-3.16.5-Linux-x86_64/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:393 (_FPHSA_FAILURE_MESSAGE)
  lib/nanopb/extra/FindNanopb.cmake:363 (find_package_handle_standard_args)
  CMakeLists.txt:45 (find_package)
```
I have grpc-tools installed and I checked that the function in proto/_utils.py returns True.

Is the grpc-tools also supported by the cmake FindNanopb?

alexpacini:
This seems to fix #560


PetteriAimonen:
Can you add `generator-bin` before the `generator` path also? That way it will probably work on Windows also.

alexpacini:
> Can you add `generator-bin` before the `generator` path also? That way it will probably work on Windows also.
Yes, I can do. Those are hints for the paths.

> Also, what does the line with only `$` do? I'm not very familiar with CMake.
Sorry typo.

In theory, it might be better to change also :
```bash
MYPATH=$(dirname "$0")
exec "$MYPATH/nanopb_generator.py" --protoc-plugin
```
to
```bash
MYPATH=$(dirname "$0")
cd ${MYPATH}
python -m nanopb_generator --protoc-plugin
```

Are you happy if I add also this change? (This should remove the ImportError with the )

PetteriAimonen:
I think there is no need for `python -m` change, as calling it directly is supposed to work and the ImportError was only an accidental bug on Python 3. The change has some potential problems for portability to systems where `/usr/bin/python` is Python 2.x, so I'd prefer to keep it as is.

alexpacini:
Ok, I will revert it, no problem. I will change the history, as this will keep it clean.

PetteriAimonen:
Looks ok to me now, thanks!

PR #567, opened 2020-08-01
More seamless SwiftPM implementation

paulb777:
This update allows clients to share common syntax for nanopb imports/includes between CocoaPods and Swift Package Manager.

See example at https://github.com/firebase/firebase-ios-sdk/pull/6165


PetteriAimonen:
Thanks!

PR #568, opened 2020-08-06
Update conanfile.py

Psy-Kai:
settings.os represent the target system not the build system (https://docs.conan.io/en/latest/systems_cross_building/cross_building.html#gnu-triplet-convention).
Thus we need to check the `settings_os_build` variable not the `os` variable.
This will fix the problem that CMake tries to check if the compiler works with wrong parameter (which results in using -`rdynamic` which may not be supported on mcu compilers)


PetteriAimonen:
Seems ok to me.

@amatosov-rbx Do you want to take a look?

PR #577, opened 2020-09-01
Better support C++ types in generated structs

otopetrik:
Option '(nanopb).callback_datatype' does not work with
most C++ types, because generated structure is inside
an 'extern "C" {' block, that prevents using namespaces, templates...

Option '(nanopb_fileopt).include' does not support including system headers,
that prevents using `<vector>`, `<string>`, etc...

Modify generator in following ways:

- start 'extern "C" {' later, after the structures

- for include option stating with `'<'`, include system header
e.g. `"(nanopb_fileopt).include = '<vector>'"` generates `"#include <vector>"`

Add 'cxx_callback_datatype', SConscript code based on 'cxx_descriptor' test.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1541,8 +1541,12 @@ class ProtoFile:
         yield '\n'
 
         for incfile in self.file_options.include:
-            yield options.genformat % incfile
-            yield '\n'
+            # allow including system headers
+            if (incfile.startswith('<')):
+                yield '#include %s\n' % incfile
+            else:
+                yield options.genformat % incfile
+                yield '\n'
 
         for incfile in includes:
             noext = os.path.splitext(incfile)[0]
@@ -1558,10 +1562,6 @@ class ProtoFile:
         yield '#error Regenerate this file with the current version of nanopb generator.\n'
         yield '#endif\n'
         yield '\n'
-
-        yield '#ifdef __cplusplus\n'
-        yield 'extern "C" {\n'
-        yield '#endif\n\n'
 
         if self.enums:
             yield '/* Enum definitions */\n'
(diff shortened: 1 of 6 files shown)


PetteriAimonen:
Looks good to me!

Regarding `"tests/site_scons/site_tools/nanopb.py:_nanopb_proto_builder forces '.c' suffix"`, I would be ok with adding e.g. "NanopbProtoCxx" builder that would apply the options required for `.cxx` extension.

PetteriAimonen:
Great, thanks!

PetteriAimonen:
Hmm, uh, looks like there is a bit of a problem: https://travis-ci.org/github/nanopb/nanopb/jobs/723668169

    build/cxx_callback_datatype/message.pb.cpp:12:1: error: offset of on non-POD type 'TestMessage' (aka '_TestMessage') [-Werror,-Winvalid-offsetof]

(Also looks like Travis CI build for pull requests wasn't working for some reason, but that should be fixed now. It would have shown the problem earlier.)

So it seems there is a bit of a conflict between C++ type rules and the way nanopb currently accesses struct members. According to https://stackoverflow.com/questions/1129894/why-cant-you-use-offsetof-on-non-pod-structures-in-c it might be safe to define a custom `pb_offsetof()` for C++ as long as virtual inheritance is not used, but I'm not sure if that is actually a good idea.

otopetrik:
> Hmm, uh, looks like there is a bit of a problem: https://travis-ci.org/github/nanopb/nanopb/jobs/723668169

Looks like clang is more strict than gcc 7.5.0
 
> So it seems there is a bit of a conflict between C++ type rules and the way nanopb currently accesses struct members. According to https://stackoverflow.com/questions/1129894/why-cant-you-use-offsetof-on-non-pod-structures-in-c it might be safe to define a custom `pb_offsetof()` for C++ as long as virtual inheritance is not used, but I'm not sure if that is actually a good idea.

Fortunately it looks like `offsetof` restrictions were relaxed in C++11.
Second attempt (#581) limits the test only for C++11 and later, and passes: https://travis-ci.com/github/nanopb/nanopb/jobs/380960911

PR #581, opened 2020-09-03
Better support C++ types in generated structs

otopetrik:
Option '(nanopb).callback_datatype' does not work with
most C++ types, because generated structure is inside
an 'extern "C" {' block, that prevents using namespaces, templates...

Option '(nanopb_fileopt).include' does not support including system headers,
that prevents using <vector>, <string>, etc...

Modify generator in following ways:

- start 'extern "C" {' later, after the structures

- for include option stating with '<', include system header
e.g. "(nanopb_fileopt).include = '<vector>'" generates "#include <vector>"

Add 'cxx_callback_datatype', SConscript code based on 'cxx_descriptor' test.

NanopbProto builder forces .pb.c extension, resulting
in a compilation failure when using C++ types for callback_datatype.

Add NanopbProtoCpp builder that generates .pb.cpp and .pb.hpp files
ensuring compilation in C++ mode (and to mark header file as C++-only).

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1541,8 +1541,12 @@ class ProtoFile:
         yield '\n'
 
         for incfile in self.file_options.include:
-            yield options.genformat % incfile
-            yield '\n'
+            # allow including system headers
+            if (incfile.startswith('<')):
+                yield '#include %s\n' % incfile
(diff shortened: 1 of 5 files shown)


PetteriAimonen:
Awesome, thanks again!

PR #584, opened 2020-09-11
Install protobuf python module

Psy-Kai:
Since the protobuf python module has to be installed on the build machine, it should be installed in the [https://docs.conan.io/en/latest/reference/conanfile/methods.html#system-requirements](system_requirements) method of the conan file.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -20,6 +20,14 @@ class NanoPbConan(ConanFile):
     def configure(self):
         if self.settings.os == "Windows" and self.settings.compiler == "Visual Studio":
             del self.options.fPIC
+            
+    def system_requirements(self):
+        import pip
+        if hasattr(pip, "main"):
+            pip.main(["install", "protobuf"])
+        else:
+            from pip._internal import main
+            main(["install", "protobuf"])
 
     def build(self):
         cmake = CMake(self)


PetteriAimonen:
What does this do if the system already has `python-protobuf` installed through package manager?

Psy-Kai:
It would install it along with the system `python-protobuf` package. But the pip version is up to date. On older systems there might not be a `python3-protobuf` package.

PetteriAimonen:
That sounds like it would cause problems if system has some version of `protoc` that is incompatible with the newest `python-protobuf` from pip. Installing `grpcio-tools` also should work, but it seems quite bloat to force them installed if there is already a perfectly usable combination installed system-wide. I wonder if there should be some check whether protobuf can be imported, and only install if it cannot be?

Another thing I want to ask: how does conan handle Python versions? Is there a possibility that conan happens to run with e.g. Python 2 and installs packages for that, and then nanopb's generator ends up running with Python 3 and cannot find them?

@amatosov-rbx Do you have an opinion on this?

Psy-Kai:
It could happen that conan runs with python2 but [conan deprecated python2](https://docs.conan.io/en/latest/installation.html#python-2-deprecation-notice) and current versions wont run with python2.
Another way of installing `python-protobuf` would be via system package manager. But I doubt that this would be a good sollution on windows (and older linux systems).
To the compatibility with `protoc`: I dont know if there would be a compatibility problem. As far as I know `protoc` calls the plugins in a separate process. If the api between the processes is stable there should not be any problem.

PetteriAimonen:
Unfortunately `protoc --python_out` is for some reason built into the protoc binary, instead of being a plugin that would come with the `python-protobuf` library. And it does have several incompatible versions. This has caused a lot of annoyance and support load in the past, that's how I know to be picky about it :)

Psy-Kai:
Why would you use `python_out` for nanopb? I didn't dive too deep into the generator script so I don't know its internals.

PetteriAimonen:
The generator script needs to communicate with `protoc`, and this communication happens using protobufs. That's why the script needs `python-protobuf`, and in order to build the necessary `.proto` files into `.py` for the generator, `protoc --python_out` is used. The code related to this is in `generator/proto` folder.

I'd love to get rid of that dependency, but I'm not sure if any of the other protobuf libraries for Python work without `protoc` either.

Psy-Kai:
And how about just adding an option like `install_protobuf_from_pip`? This way the consumer of the conanfile can choose between manually installing the proto python libs and using the one from pip.

PetteriAimonen:
Having an option sounds good to me :)

But I suggest if you make it install `protobuf` from pip, it should also install `grpcio-tools` from pip. That package provides `protoc`.

Psy-Kai:
Conan expects that compilers get installed separately. It would be very strange if you create a project using nanopb (for MCU Targets) and googles protobuf (for x86 Targets) and protoc gets installed by nanopb. This would imply that you rely on nanopb to install protoc. And if googles protobuf also installs protoc from another source you get the very same problem conan tries to avoid: multiple binaries of one thing gets installed.

So I would just install `protobuf` from pip and not `grpcio-tools`. The user has to install its compilers manually (like the user has to install gcc/clang/msvc manually).

PetteriAimonen:
So what does it help then to install `python3-protobuf` automatically, when user has to go find a matching `protoc` manually anyway? Sounds like it would help only in the weird situation where a distribution provides a new enough `protoc`, but does not provide `python3-protobuf`.

Have you tested the proposed functionality on some cleanly installed distro/docker?

Psy-Kai:
Could it be that the problem was that someone wanted to mix proto2 and proto3?
Using protoc version 3.6.0 with pip3 protobuf 3.13.0 does not lead to an error:
```
FROM conanio/gcc48

RUN pip3 install protobuf && \
    git clone https://github.com/protocolbuffers/protobuf.git && cd protobuf && \
    ./autogen.sh && ./configure && make -j10 && sudo make install && sudo ldconfig && cd .. && \
    \
    git clone https://github.com/nanopb/nanopb.git && cd nanopb && \
    cd ..
    
RUN echo 'syntax="proto3";\npackage hello.world;\n\nmessage Hello {\n  message World {\n    int32 value = 1;\n  }\n  World world = 1;\n}' > helloworld.proto && \
    protoc --plugin=protoc-gen-nanopb=nanopb/generator/protoc-gen-nanopb helloworld.proto --nanopb_out=.
```

I used conans gcc4.8 image because gcc 4.8 would be the oldest compiler I would use. And it is based on an old version of ubuntu. Newer versions justs work as well.

Psy-Kai:
Could it be that the problem was that someone wanted to mix proto2 and proto3?
Using protoc version 3.6.0 with pip3 protobuf 3.13.0 does not lead to an error:
```
FROM conanio/gcc48

ENV PROTOBUF_VERSION=3.6.0

RUN pip3 install protobuf && \
    wget https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOBUF_VERSION/protobuf-cpp-$PROTOBUF_VERSION.tar.gz && \
    tar -xvzf protobuf-cpp-$PROTOBUF_VERSION.tar.gz && cd protobuf-$PROTOBUF_VERSION && \
    ./configure && make && sudo make install && sudo ldconfig && cd .. && \
    \
    git clone https://github.com/nanopb/nanopb.git
    
RUN echo 'syntax="proto3";\npackage hello.world;\n\nmessage Hello {\n  message World {\n    int32 value = 1;\n  }\n  World world = 1;\n}' > helloworld.proto && \
    protoc --plugin=protoc-gen-nanopb=nanopb/generator/protoc-gen-nanopb helloworld.proto --nanopb_out=.
```
I used conans gcc4.8 image because gcc 4.8 would be the oldest compiler I would use. And it is based on an old version of ubuntu. Newer versions justs work as well.

PetteriAimonen:
Yeah, it seems we enjoyed a long span of stability in python-protobuf. But there was another breaking change this month: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates

So I still think it causes more harm than good to unconditionally install `python-protobuf` from pip.

PR #596, opened 2020-09-30
Allow multiple targets to be created by calling NANOPB_GENERATE_CPP()

leabut:
When calling NANOPB_GENERATE_CPP() multiple times with cmake ninja generator, one will run into following issue:
```
ninja: error: build.ninja:14698: multiple rules generate nanopb/generator/proto/nanopb_pb2.py [-w dupbuild=err
```

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -149,7 +149,7 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
 
   list(REMOVE_DUPLICATES _nanopb_include_path)
 
-  set(GENERATOR_PATH ${CMAKE_BINARY_DIR}/nanopb/generator)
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Looks ok to me, but I don't understand much about CMake.

@alexpacini, @marekzukal, @efisher Do you want to take a look?

leabut:
The changes introduced will allow to copy generator relevant files (like shown above, nanopb_pb2.py) to the build directory,
where NANOPB_GENERATE_CPP() was called from.

The problem kind of still remains if calling NANOPB_GENERATE_CPP() cmake function multiple times from same binary directory.
But for general use cases one should have different binary directories for different sw modules calling cmake functions, therefore it will get pretty unlikely to run into the same issue again.

twam:
Sounds to me like a reasonable change.

PetteriAimonen:
Thanks :)

Issue #603, opened 2020-10-08
nanopb_generator does not report a version number


mrkita:
I can't find any sort of version number reporter for the python generator (at least in the compiled binaries provided in the website).

This would be very helpful for provisioning systems, that can use that to verify the installed version number.

PetteriAimonen:
Yeah, it would be useful to add a `--version` switch.

In the meanwhile, one can use this snippet: `python -c 'import nanopb_generator; print(nanopb_generator.nanopb_version)'`

mrkita:
Thanks that's a good suggestion. The problem is, we don't have python installed directly, nor do we want to include it as a dependency (nothing else in our project makes use of it).

tknerr:
+1, a `--version` flag would definitely be helpful 👍 :)

PR #607, opened 2020-10-16
Add '--version' option to nanopb_generator.py

WakandaO2:
As requested in #603 ,

New option "--version" was added to nanopb_generator.py that will print nanopb's version and exit.

This option is supported in both CLI and protoc-plugin mode.


PetteriAimonen:
Thanks!

PR #645, opened 2021-03-09
Add support for source comments in generated headers

AngelVI13:
Add support for proto file comments in generated headers.

Supports*:

- [X] Enum member comments (leading & trailing)
- [x] Enum comment (leading & trailing)
- [x] Message member comments (leading & trailing)
- [x] Message comment (leading & trailing)

\* Planning to implement all of the above

In order to enable this feature `protoc` must be called with `--include_source_info` flag. If that is not the case - these changes do not have any effect on generated headers.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1608,10 +1608,10 @@ class ProtoFile:
                     field.type_name = mangle_field_typename(field.type_name)
 
             self.messages.append(Message(name, message, message_options))
-            for enum in message.enum_type:
+            for index, enum in enumerate(message.enum_type):
                 name = create_name(names + enum.name)
                 enum_options = get_nanopb_suboptions(enum, message_options, name)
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Looks good so far!

AngelVI13:
Sorry about all the "fix" commits but for some reason I can't seem to run the tests locally (scons fails with "plugin error 127"). I will investigate further. 

If everything is okay regarding the implementation approach then I will continue with the next few checkpoints. Do you have any requests or concerns regarding those ?

PetteriAimonen:
No problem about the fixes, you can do `git rebase -i HEAD~20` and `git push -f` when everything is done to clean them up, or I can do it at the merge time.

The implementation direction looks good.

PetteriAimonen:
Thanks a lot for the great work! This is merged now.

PR #648, opened 2021-03-20
 Fix issue 637

tymmej:
Cache union and size definitions before writing it to file. Group
 entries under same cpp_guard in dictionary. Duplicated entries will be
 ignored due to set ignoring new entries. After that convert set to
 list and sort it. It prevents mess when regenerating files.


PetteriAimonen:
Thanks, I'll take a look.

PetteriAimonen:
Merged in ea10622 with small modifications.

PR #652, opened 2021-03-29
Fixing uninitialized PYTHON_INSTDIR in CMakeLists.txt

odolgy:
The variable PYTHON_INSTDIR is not initialized if the option
nanopb_BUILD_GENERATOR is OFF


PetteriAimonen:
Thanks!

PR #674, opened 2021-06-03
Fix Xcode 12.5 analyze issue

### LINKED ISSUE #667: Xcode 12.5 analyzer shows two warnings in pb_decode.c
From https://stackoverflow.com/questions/67644723/updated-firebase-crashlytics-on-ios-getting-a-bunch-of-analyzer-issues

```
/Users/paul/Library/Developer/Xcode/DerivedData/firebase-ios-sdk-ayfzngrufgjzprffkljwxevrtomi/SourcePackages/checkouts/nanopb/pb_decode.c:541:24: warning: Dereference of null pointer [core.NullDereference]
        *(void**)pItem = NULL;
        ~~~~~~~~~~~~~~~^~~~~~
/Users/paul/Library/Developer/Xcode/DerivedData/firebase-ios-sdk-ayfzngrufgjzprffkljwxevrtomi/SourcePackages/checkouts/nanopb/pb_decode.c:547:9: warning: Null pointer passed as 1st argument to memory set function [unix.cstring.NullArg]
        memset(pItem, 0, iter->pos->data_size);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

paulb777:
Fix #667

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -639,7 +639,9 @@ static bool checkreturn decode_pointer_field(pb_istream_t *stream, pb_wire_type_
 
                     /* Decode the array entry */
                     pItem = *(char**)iter->pData + iter->pos->data_size * (*size);
-                    initialize_pointer_field(pItem, iter);
+                    if (pItem != NULL) {
+                        initialize_pointer_field(pItem, iter);
+                    }
                     if (!func(&substream, iter->pos, pItem))
                     {
                         status = false;


PetteriAimonen:
Hmm, this change is semantically a bit weird, because it skips the initialization but calls `func()` anyway. Of course there is no difference as long as `pData` is not `NULL`, which it shouldn't be at this point, but it just looks weird when reading the code.

I committed a slightly modified version, can you check if it fixes the warning?

paulb777:
Yes. Confirmed that https://github.com/nanopb/nanopb/commit/b6c6381949ed3f7101c95c79beee66a6560e2a10 fixes the warning. Thanks!

BTW, if I'm correctly interpreting the 41 steps the analyzer traces back to think it can be a NULL, both iter->pData and *size can come from unknown input data.

Screen Shot 2021-06-04 at 8 53 01 AM

PetteriAimonen:
Thanks for the screenshot, it clarifies what confuses the analyzer: on line 622, it assumes the condition is false. But actually `*size + 1` is always larger than `allocated_size` on first iteration, because `allocated_size` is initialized to `*size`.

This part of the code is largely same in 0.4 branch, so I committed the same fix there.

PR #669, opened 2021-05-31
[CMake] add build interface

mathisloge:
If you add nanopb as a submodule. The headers won't be available correctly when building other projects in the superbuild.

Therefore I've added the `$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>`

PR #676, opened 2021-06-05
Unify CMake

mathisloge:
this PR is a followup of #669

* Canges the raw `nanopb_PROTOC_PATH` with a find_program approch which is more the "cmake way to do".
* Adds the missing build interface for the shared build
* Adds component Interpreter to find python since only the interpreter is needed
* Removes some whitespaces to get a unified form.


mathisloge:
\ping @mrostecki since you was marked at the previous PR :)

mrostecki (approved):
looks good to me

PR #677, opened 2021-06-08
Use CMAKE inbuilt mechanic to decide if protoc was found

zrno:
-The old way was wrong and would report errors when protoc was found. The correct way to do it manually would be to use if(NOT EXISTS ${nanopb_PROTOC_PATH})

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,10 @@ option(nanopb_BUILD_RUNTIME "Build the headers and libraries needed at runtime" 
 option(nanopb_BUILD_GENERATOR "Build the protoc plugin for code generation" ON)
 option(nanopb_MSVC_STATIC_RUNTIME "Link static runtime libraries" ON)
 
-find_program(nanopb_PROTOC_PATH protoc REQUIRED)
+find_program(nanopb_PROTOC_PATH protoc)
+if(NOT EXISTS ${nanopb_PROTOC_PATH})
+    message(FATAL_ERROR "protoc compiler not found")
+endif()
 
 if(NOT DEFINED CMAKE_DEBUG_POSTFIX)
     set(CMAKE_DEBUG_POSTFIX "d")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -14,10 +14,7 @@ option(nanopb_BUILD_RUNTIME "Build the headers and libraries needed at runtime"
 option(nanopb_BUILD_GENERATOR "Build the protoc plugin for code generation" ON)
 option(nanopb_MSVC_STATIC_RUNTIME "Link static runtime libraries" ON)
 
-find_program(nanopb_PROTOC_PATH protoc)
-if(NOT EXISTS nanopb_PROTOC_PATH)
-    message(FATAL_ERROR "protoc compiler not found")
-endif()
+find_program(nanopb_PROTOC_PATH protoc REQUIRED)
```
zrno, line 17: I see there is a protoc binary bundled with nanopb. Should I add a path to that binary in case protoc is not found on the users machine?

PetteriAimonen: Hmm yeah, it would be good to have `generator-bin` first in the search path if it exists. Presumably those who download a binary package want to use the generator that comes with it.

PetteriAimonen:
@mathisloge Just pinging in case you want to take a look.

mathisloge:
Required isn't compitable with the specified cmake minimum version of 2.8. That's the reason why I haven't used required

mathisloge:
But maybe we should replace the if condition with `not [proto var]` if it causes issues

zrno:
Made a change to be compatible with CMake 2.8
Any reason not to update to a more recent CMake version?

CMakeLists.txt
zrno: Do you want me to add a hint to CMake to search in `nanopb/generator/` if protoc was not found in the default paths on the user machine?

PetteriAimonen: Yeah, that could be last in the path. The `protoc` script in `generator` folder is just a wrapper around `grpcio-tools` Python package.

PR #678, opened 2021-06-08
Use the right path when checking for protoc path

cerna:
This is a small PR which addresses a bug introduced in #676 when checking for the `protoc` generator/compiler executable:

#676 started testing for file or directory existence via the integrated CMake function `if(EXISTS)`in the form `if(NOT EXISTS nanopb_PROTOC_PATH)`, however for the CMake to expand the variable, the use of `${}` needed, otherwise it will try to test for existence of literary the `nanopb_PROTOC_PATH`as a path.


cerna:
Sorry, haven't looked at #677 first.

Issue #589, opened 2020-09-22
Problem with IAR Compiler; fields struct not generated and 64 Bit Types possibly not supported


krisnov1:
Hi folks,
I have a problem.
 I started generation C code without a problem.
The first compilation problem occured here: pb.h

**If you get errors here, it probably means that your stdint.h is not corect for your platform.
#ifndef PB_WITHOUT_64BIT
PB_STATIC_ASSERT(sizeof(int64_t) == 2 * sizeof(int32_t), INT64_T_WRONG_SIZE)
PB_STAT**
```
I fixed that by using the switch **PB_WITHOUT_64BIT**. Hope that is OK

The second compilation problem occured here: simple.pb.c 

Error[Pe077]: this declaration has no storage class or type specifier C:\Users\OXF_KNO\work\blueprint-package-integration-test\tpt_test\Nucleo_Node\Nucleo_Cfg\simple.pb.c 9 
"PB_BIND(SimpleMessage, SimpleMessage, AUTO)"


I figured out the structure was not generated. Don’t know why
I was abel to “fix” that problem by using the switch **#define PB_NO_STATIC_ASSERT ////TODO KN?**
 


  buffer = (8, 1, 16, 2, 24, 3, 145, 1, 24, 180, 164, 224, 179, 5, 0, 0, 0, 0, 0,
  message = (
    lucky_number1 = 0x1,
    lucky_number2 = 0x2,
    lucky_number3 = 0x3)

 
The data gets written and retrieved as expected. BUT the pb_decode returns an error status.

What am I doing wrong here..
Regards

Krisitjan

PetteriAimonen:
It does sound a lot like the compiler or it's preprocessor is not standard-compliant enough to handle the macro tricks that nanopb uses to generate the descriptor for the structures.

Are you able to upload the preprocessed output of the .pb.h file and corresponding .pb.h?

krisnov1:
[List.zip](https://github.com/nanopb/nanopb/files/5266482/List.zip)
The seems to be no pb.h (preprocessed output ). But I found structures in simple.pb.i. I send all the preprocessed output to make sure you are not missing something .

PetteriAimonen:
Hmm, it sounds like the main problem could be that `PB_STATIC_ASSERT` macro is not working correctly on IAR for some reason. I don't have that compiler myself, so it is pretty difficult to troubleshoot.

[...]

PR #679, opened 2021-06-15
Use 'static_assert'  keyword for iar

ghseb:
IAR compiler provides a `static_assert` keyword that can be used to prevent a compiler error due to `_Static_assert` not being available.

Related: #589


PetteriAimonen:
Thanks!

PR #681, opened 2021-06-22
Add clean target

ghseb:
I'm not sure how the Makefile is used in other contexts, but for our use case need to clean the generated python file explicitly some times.


PetteriAimonen:
Makes sense.

Issue #692, opened 2021-07-26
Duplicate declarations of size_unions with repeated fields inside a oneof


IngoKresse:
This seems to be another incarnation of #637, tested against version 0.4.5 and master. Most of the investigations and the guards were done by my colleague Henrik Poppe (@AlmostAll), thanks a lot!

If one part of a oneof is a message in another file which has a repeated field (with a max_count option), then the declaration of the size_union may be emitted multiple times. Here are example messages that reproduce the problem:

a.proto
```protobuf
syntax = "proto3";

import "b.proto";

package a;

message A {
  oneof data {
     int32 number = 1;
     b.BigList1 message = 2;      // unknown max_count at this point (?) -> generates size_union
  }
}

message AA {
  b.BigList2 another_message = 2; // second sub-message of unknown size -> causes second guard block
  A sub_message = 1;              // here, the size_union gets duplicated :-(
}
```

b.proto
```protobuf
syntax = "proto3";

package b;

message BigList1 {
  repeated double numbers = 1;
}

message BigList2 {
  repeated double more_numbers = 1;
}
```

b.options
```
b.BigList1.numbers max_count: 5;
b.BigList2.more_numbers max_count: 6;
```

**What happens?**

The following code is generated inside a.pb.h:

```cpp
/* Maximum encoded size of messages (where known) */
#if defined(b_BigList1_size)
#define a_A_size                                 (0 + sizeof(union a_A_data_size_union))
union a_A_data_size_union {char f2[(6 + b_BigList1_size)]; char f0[11];};
#endif
#if defined(b_BigList1_size) && defined(b_BigList2_size)
#define a_AA_size                                (12 + sizeof(union a_A_data_size_union) + b_BigList2_size)
union a_A_data_size_union {char f2[(6 + b_BigList1_size)]; char f0[11];};
#endif
```

The union `a_A_data_size_union` is declared twice and the compiler throws an error.

**What should happen?**

I see several possibilities  how this could be solved:

1. If the generator could read the options for the dependencies, then it may be able to compute the maximum sizes directly.
2. The generator could only emit the size_union declaration together with the message where the oneof is defined, in a patch like this:
```diff
--- ./old/nanopb_generator.py<->2021-03-22 14:44:15.000000000 +0000
+++ ./new/nanopb_generator.py<->2021-07-23 13:14:12.988435713 +0000
@@ -1811,7 +1811,13 @@
                     if cpp_guard not in guards:
                         guards[cpp_guard] = set()
[...]

PR #693, opened 2021-07-26
Emit size_union only once, with guards

IngoKresse:
This Pull-Request fixes issue #692, where the same sizeof_union may get generated twice. This change only emits a sizeof_union once per file.


PetteriAimonen:
Thanks!

This will need a regression test also. It could be similar as `tests/regression/issue_637`.
If you have time to add one, it would be great; otherwise I'll add one next week.

PetteriAimonen:
Thanks, this is now merged as commit 863b7390b774313e4e7ab6a42c31bf0bfe846023

PR #718, opened 2021-12-01
Added generator support for PlatformIO

Hacker-CB:


PetteriAimonen:
Thanks.

I wonder if automatic test for this could be added somehow. Preferrably as a GitHub CI task or similar.

Hacker-CB:
Ok, I will add test

Hacker-CB:
@PetteriAimonen where to put test platformio project?

I think `tests` folder is not good place beacuse it has specific structure?
I can make something like `tests_extra` folder for this case, or could you point me to better place?

PetteriAimonen:
If it is possible to do as a Github workflow, `.github/workflows` would be the correct place.

The `tests` folder is indeed mainly for unit tests, while this is more like an integration test.

Hacker-CB:
I know about workflows, but tests require some specific C-code with `hello world` and `ini file`. 
So, seems that `workflows` folder is not good place for that.
I think it will be better to put to `workflows` only some commands which run actual test.

Workflow plan:

1. Setup platformio
2. Run platformio build of the some test project(s).

So, I need some place inside repository to put test projects for step 2.

PetteriAimonen:
Could the test project go into `examples/platformio`?

CMake builds are currently tested in a similar way by building `examples/cmake_simple`.

Hacker-CB:
> Could the test project go into `examples/platformio`?

Ok. I will do in this way. 
Thank you.

Hacker-CB:
@PetteriAimonen all tests done

PetteriAimonen:
Awesome! Thanks a lot!

PR #720, opened 2021-12-09
Parse all include files before generating code

inolen:
When generating code for proto files which used includes, unless all of the includes where passed as inputs I would get an error for types defined in imported files:

```
Could not find enum type %s while generating default values for %s.
Try passing all source files to generator at once, or use -I option.
```

It looks like this was possibly unintentionally introduced with 0bd67e3dea73e03d2b85c373a73ef1879774a5fe. This PR re-enables parsing imports before generating code, with the difference being that it only generates code for the filenames passed as inputs, not also for the files they imported.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -2216,8 +2216,9 @@ def main_cli():
                          % (google.protobuf.__file__, google.protobuf.__version__))
 
     # Load .pb files into memory and compile any .proto files.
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks! Sometimes I wish I could ask my past self questions, though I fear the answer would be just "I wanted to get it fixed".

There should be more tests covering cases like this, but that can wait for later.

PR #723, opened 2021-12-11
Add flag to disable requiring protoc as part of PlatformIO build

devalbo:
Hi - I love using nanopb with Arduino and PlatformIO! I added a dependency on the nanopb library to a PlatformIO project and I am getting the message `[nanopb] No `nanopb_protos` specified, exiting.` when I run `pio run`.

I already have my nanopb header and `.c` files in place and am using the library dependency for `pb.h' and the `pb_*code.h/c` files. It would be nice if there were an option like the attached `nanopb_require_protoc` build flag to bypass the library using its own `protoc` step.

### DIFF
--- a/generator/platformio_generator.py
+++ b/generator/platformio_generator.py
@@ -5,123 +5,133 @@ from platformio import fs
 
 Import("env")
 
-nanopb_root = os.path.join(os.getcwd(), '..')
 
-project_dir = env.subst("$PROJECT_DIR")
-build_dir = env.subst("$BUILD_DIR")
+def do_platformio_protoc():
+    nanopb_root = os.path.join(os.getcwd(), '..')
 
-generated_src_dir = os.path.join(build_dir, 'nanopb', 'generated-src')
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
@hacker-cb What do you think?

hacker-cb:
Its normal behaviour. 
It's just information message, not error. Exit code is `0`.

`nanopb_protos` not specified:

https://github.com/nanopb/nanopb/blob/3d65777c1379dc6ce329779f1294b8c7638d1fcc/generator/platformio_generator.py#L21

`nanopb_protos` specified, but no files matched regexp - error code `1`: https://github.com/nanopb/nanopb/blob/3d65777c1379dc6ce329779f1294b8c7638d1fcc/generator/platformio_generator.py#L31

@PetteriAimonen we can rename `[nanopb] No nanopb_protos specified, exiting` message to other one, may be like `[nanopb] No nanopb_protos: nothing to do` or similar else?

PetteriAimonen:
Yeah, that would be a clearer message.

PR #727, opened 2021-12-14
Remove dependency on Python distutils

ianjfrosst:
As of Python 3.10, the `distutils` module is deprecated ([PEP 632](https://www.python.org/dev/peps/pep-0632/)), and importing it triggers a warning.

This PR removes all references to `distutils` from nanopb.


PetteriAimonen:
I think this should be ok and I'll merge it.

@zrno @mathisloge @odolgy @dflogeras If some of you want to try out the CMakeLists changes to verify that they work on your system also, that would be great!

mathisloge:
Works on my machine

zrno:
Works on my machine also

PR #736, opened 2021-12-24
PlatformIO: Fix config parser warning

hacker-cb:


hacker-cb:
@PetteriAimonen please keep in mind this will break exsiting platformio configs, which use generator with old `nanopb_protos`/`nanopb_options`.

PetteriAimonen:
Yeah - fortunately it is pretty new feature so breaking at this point should be ok still.

Thanks for all the work you have done on platformio support!

PR #751, opened 2022-02-11
Use memcpy instead of iterating on buf_read/write

gil-shapira-intusurg:
**Description:**
Enabling `PB_BUFFER_ONLY` should speed up execution ([ref](https://github.com/nanopb/nanopb/blob/1d43d109af32ef2b083ba0312259e2e06407f023/docs/reference.md))
When the flag is set, the `buf_read`/`buf_write` functions are used to copy data from/to a stream respectively.
Using iteration is much inferior to `memcpy`.
In my case, switching to `memcpy` increased the performance by a factor of 8.

### DIFF
--- a/pb_decode.c
+++ b/pb_decode.c
@@ -67,14 +67,12 @@ typedef struct {
 
 static bool checkreturn buf_read(pb_istream_t *stream, pb_byte_t *buf, size_t count)
 {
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Makes sense, thanks!

PR #756, opened 2022-02-21
CMake fix: keep full filename without last extension for generated files

matschbirne:
I wanted to generate the following .proto files with CMake:

- `foo.something.proto`
- `foo.somethingelse.proto`

Since CMake took the filename without the longest extension (NAME_WE) for the generated C files, the resulting name for both files would be `foo.pb.c` (and `foo.pb.h`) which caused the following error:

```
CMake Error at XXX/nanopb/extra/FindNanopb.cmake:276 (add_custom_command):
  Attempt to add a custom rule to output

    YYY/protocols/foo.pb.c.rule

  which already has a custom rule.
Call Stack (most recent call first):
  CMakeLists.txt:30 (nanopb_generate_cpp)
```

This PR fixes this issue for me by taking the filename without only the last file extension (NAME_WLE). The resulting names of the generated C files are now `foo.something.pb.c` and  `foo.somethingelse.pb.c`, which is fine. 

I hope this PR doesn't cause any other issues. Feel free to give me some feedback.


PetteriAimonen:
Thanks!

PR #764, opened 2022-03-15
PYTHON_INSTDIR is not defined if nanopb_BUILD_GENERATOR=OFF

avikde:
When using nanopb with add_subdirectory and setting this option to off, this install line fails because the PYTHON_INSTDIR variable is empty.


PetteriAimonen:
Thanks!

Issue #772, opened 2022-04-09
new (global) option like "FT_DEFAULT" but fallback to FT_POINTER instead of FT_CALLBACK 


ConradWood:
Hi,

I propose to introduce at least one extra option to add specific functionality.
The functionality I like to introduce is to generate fields as static fields, if possible, but fields with variable length to be allocated dynamically via malloc (realloc).
Currently, the behaviour of FT_DEFAULT is to generate static fields if it can and generate callbacks if it cannot create static fields.
I propose to keep FT_DEFAULT behaviour "as-is" for now.

Instead, I propose to create a new option, called "FT_AUTO_FB_POINTER". 
For example: "nanopb_generator.py -s type=FT_AUTO_FB_POINTER [src]".

The behaviour of the options shall be identical to FT_DEFAULT, except, it would use pointers and malloc to generate dynamic fields.

in "nanopb_generator.py" I see:

  if field_options.type == nanopb_pb2.FT_DEFAULT:
            if can_be_static:
                field_options.type = nanopb_pb2.FT_STATIC
            else:
                field_options.type = nanopb_pb2.FT_CALLBACK

so IMHO the change is fairly trivial. We would *add* below:

  if field_options.type == nanopb_pb2.FT_AUTO_FB_POINTER:
            if can_be_static:
                field_options.type = nanopb_pb2.FT_STATIC
            else:
                field_options.type = nanopb_pb2.FT_POINTER

(of course, also modifying the options.proto files for enums).

Note that the naming scheme "FT_AUTO_FB_POINTER" is intended to be extensible.
Perhaps someone, some day, wants to add FT_AUTO_FB_POINTER_FOR_STRING_AND_CALLBACK_FOR_ARRAY
(the example might be a little long, but it's for illustrative purposes only :) )
(FB stands for "Fallback").
With that naming scheme, FT_DEFAULT would eventually become FT_AUTO_FB_CALLBACK.

Would a patch to that extent be welcome?

Conrad

PetteriAimonen:
I think it might be better to have a separate `optional FieldType fallback_type = x [default = FT_CALLBACK]` option in `NanoPBOptions`. This would allow fallback to `FT_IGNORE` also.

With regards to specific field types based on datatype, I think we could eventually have more flexible selectors in `.options` file. Something like `[datatype:TYPE_BYTES] * type:FT_POINTER`

ConradWood:
[...]

PR #773, opened 2022-04-10
make the default fallback strategy for FT_DEFAULT configurable

ConradWood:
As discussed here:

https://github.com/nanopb/nanopb/issues/772#issuecomment-1094100183

Note: testsuite ran and completed successfully.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -488,7 +488,7 @@ class Field:
         self.callback_datatype = field_options.callback_datatype
         self.math_include_required = False
         self.sort_by_tag = field_options.sort_by_tag
-
+        
         if field_options.type == nanopb_pb2.FT_INLINE:
             # Before nanopb-0.3.8, fixed length bytes arrays were specified
             # by setting type to FT_INLINE. But to handle pointer typed fields,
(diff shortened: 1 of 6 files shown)


PetteriAimonen:
Thanks! Are you interested in adding a test case for this also? It could go in `tests/options`.

ConradWood:
tbh not keen - python isn't my first or preferred environment.
But I see the value and need, so I will have a go at it.

PetteriAimonen:
Thanks, if you get stuck just let me know and I'll see about it. Should only need to add usage tests of the new option to `options.proto` and then in `options_h.expected` the pattern in `.pb.h` file that should be checked for.

ConradWood:
I added the test case for fallback_type now.
Would you kindly consider it for inclusion again?

ConradWood:
The scons version on debian (my environment) passed the test, but not on the ubuntu workflow. I believe the latest patch fixes this, albeit I did not verify this on ubuntu. Please trigger the workflow again

PetteriAimonen:
Thanks!

ConradWood:
Thank you!

On 13 April 2022 10:00:05 WEST, Petteri Aimonen ***@***.***> wrote:
>Thanks!
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/nanopb/nanopb/pull/773#issuecomment-1097745171
>You are receiving this because you authored the thread.
>
>Message ID: ***@***.***>
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #775, opened 2022-04-21
custom 'realloc' with a context


ConradWood:
Hi,

In our firmware is a concept of 'criticality' of different sections. The firmware offers functionality via an API and manages partitions for different binary sections to execute. So, on each module might be 2 or more different sections running. For example, a 'critical', real-time machine controlling section and a 'user display', not-quite-as-critical section.
Any section can call the firmware with a pre-assigned handle. The kernel then prioritises resources depending on what's critical and what is not. 
For nanopb, we'd like to use different 'realloc' strategies and ram areas, depending on the section calling "pb_decode()". 
I'm pondering how to do that.
My current thinking is to add a new function 
bool pb_decode_ex2(pb_istream_t *stream, const pb_msgdesc_t *fields, void *dest_struct, unsigned int flags,void *opaque_context);

Then I would pass the opaque_context to pb_decode_inner and along the functions all the way to custom_realloc().

I would then define pb_realloc as:
#define pb_realloc(ptr, size,opaque_context) custom_realloc(ptr, size,opaque_context)

the default definition would change to:
#define pb_realloc(ptr, size,opaque_context) realloc(ptr, size)

The API would remain backwards compatible. 
This proposal is is also the least intrusive way I can come up with OTH.
It would only deployments of nanopb if a) PB_ENABLE_MALLOC is true AND b) a custom realloc is supplied

Stack usage might increase by a few bytes to push the extra pointer on the stack.

I'm happy to do a patch if that patch is likely to be accepted.

Conrad

PetteriAimonen:
I agree with your use case.

My thoughts would be:

1. `pb_istream_t` is already used as a kind of context for decoding, as it stores also the error status. It is already available in most functions calling `reallocate_field()`. Notably it is not available in `pb_release()`.
2. Changing `pb_realloc` definition would break existing applications that redefine it. An alternative would be to define a new `pb_realloc_ctx()` that by default calls `pb_realloc()`.

On the other hand, without changing nanopb code, you could use thread-local variables to pass information to the custom realloc implementation.

ConradWood:
On Wed, 2022-04-20 at 22:40 -0700, Petteri Aimonen wrote:
> I agree with your use case.
> My thoughts would be:
>    1. pb_istream_t is already used as a kind of context for decoding,
> as it stores also the error status. It is already available in most
> functions calling reallocate_field(). Notably it is not available in
> pb_release().

-v please :) 
[...]

PR #779, opened 2022-05-09
fix for custom 'realloc' with a context #775

ConradWood:
implements code and tests as discussed here https://github.com/nanopb/nanopb/issues/775
also patches example 'simple' to compile unconditionally (.PHONY) (because it seems useful for an example to reuse latest and it compiles in less than a second anyways) and with malloc support.

### DIFF
--- a/pb_decode.h
+++ b/pb_decode.h
@@ -43,6 +43,10 @@ struct pb_istream_s
 #ifndef PB_NO_ERRMSG
     const char *errmsg;
 #endif
+  #ifdef PB_ENABLE_MALLOC_CONTEXT
+  /* Note: we do not need a corresponding free() here, because we can (and do) call realloc with size==0 to free memory. */
+  void *(*realloc)(pb_istream_t *stream,void *ptr, size_t size);
+  #endif
 };
 
 #ifndef PB_NO_ERRMSG
@@ -107,7 +111,7 @@ bool pb_decode_ex(pb_istream_t *stream, const pb_msgdesc_t *fields, void *dest_s
 #define pb_decode_delimited_noinit(s,f,d) pb_decode_ex(s,f,d, PB_DECODE_DELIMITED | PB_DECODE_NOINIT)
 #define pb_decode_nullterminated(s,f,d) pb_decode_ex(s,f,d, PB_DECODE_NULLTERMINATED)
 
-#ifdef PB_ENABLE_MALLOC
+#if defined(PB_ENABLE_MALLOC) || defined(PB_ENABLE_MALLOC_CONTEXT)
 /* Release any allocated pointer fields. If you use dynamic allocation, you should
  * call this for any successfully decoded message when you are done with it. If
  * pb_decode() returns with an error, the message is already released.
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -11,6 +11,7 @@ The following options can be specified in one of two ways:
 includes nanopb headers.
 
 * `PB_ENABLE_MALLOC`: Enable dynamic allocation support in the decoder.
+* `PB_ENABLE_MALLOC_CONTEXT`: Enable dynamic allocation support in the decoder using a function pointer in istream instead of library function realloc()
 * `PB_MAX_REQUIRED_FIELDS`: Maximum number of proto2 `required` fields to check for presence. Default value is 64. Compiler warning will tell if you need this.
 * `PB_FIELD_32BIT`: Add support for field tag numbers over 65535, fields larger than 64 kiB and arrays larger than 65535 entries. Compiler warning will tell if you need this.
 * `PB_NO_ERRMSG`: Disable error message support to save code size. Only error information is the `true`/`false` return value.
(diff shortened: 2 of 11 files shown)


PetteriAimonen:
The implementation looks mostly good.
I have a few notes:

1. The comments indicate that `realloc()` would be used for releasing, but I don't see the corresponding modification to `pb_release_single_field()`. It also looks like the test case does not check for releasing through the custom allocator, it could count the number of new allocations (`ptr == NULL`) and number of releases (`size == 0`).
2. I don't see the point of changing `examples/simple` Makefile. That example is supposed to be basic and simple, and doesn't use dynamic allocation in any case.
3. I would like the `PB_ENABLE_MALLOC` to remain the main and only switch for dynamic allocation. I.e. if it is not specified, `PB_ENABLE_MALLOC_CONTEXT` should add the extra field in `pb_istream_t`, but not enable dynamic allocation. I think in the next major release when breaking changes are made, I might make `PB_ENABLE_MALLOC_CONTEXT` default on, but `PB_ENABLE_MALLOC` will remain default off.
4. The `alltypes` style test cases are a bit of a maintenance burden. I'll probably merge `custom_realloc` to `alltypes_pointer` test case at some point. But I guess it is reasonable to have it separate for now when `PB_ENABLE_MALLOC_CONTEXT` remains a default-off feature.

ConradWood:
On Mon, 2022-05-09 at 10:47 -0700, Petteri Aimonen wrote:
> The implementation looks mostly good.
> I have a few notes:
>    1. The comments indicate that realloc() would be used for
> releasing, but I don't see the corresponding modification to
> pb_release_single_field(). It also looks like the test case does not
> check for releasing through the custom allocator, it could count the
> number of new allocations (ptr == NULL) and number of releases (size
> == 0).

Yes, this is where it gets interesting. As we discussed here[1] we
discussed using a separate context, but eventually settled on using
istream instead.
Whilst implementing this is, I realised that istream is not currently
passed to pb_release().
I also realised, that my usecase is not limited to one proto, I am
thinking along this:

struct context *ctx = pb_createcontext()
pb_decode(ctx,stream_a,fields_a,dest_a);
pb_decode(ctx,stream_b,fields_b,dest_b);
ctx->close()

where memory allocations for both proto parsers come from one memory
heap. close releases both.
pb_release(ctx) is optional in this case.

It seems less than ideal to use the stream for these cases.
However, there are a number of ways I could do this with the stream:

1. I could add an api pb_release_context(pb_istream_t *stream,const
pb_msgdesc_t *fields, void *dest_struct)

2. I could modify "struct pb_msgdesc_s". Specifically I could add a
field which could hold a copy of the pointer to realloc from the
stream(). (I do not particularly like this option, for various reasons
I am happy to elaborate if needed)

My conclusion though is, that none of the options are great. I think
the original concept of the context being separate from the stream is
logically more in-line with what I (and perhaps others) would expect.

wdyt?


[1]
https://github.com/nanopb/nanopb/issues/775




>    2. I don't see the point of changing examples/simple Makefile.
> That example is supposed to be basic and simple, and doesn't use
> dynamic allocation in any case.

ok

>    3. I would like the PB_ENABLE_MALLOC to remain the main and only
> switch for dynamic allocation. I.e. if it is not specified,
> PB_ENABLE_MALLOC_CONTEXT should add the extra field in pb_istream_t,
> but not enable dynamic allocation. I think in the next major release
> when breaking changes are made, I might make PB_ENABLE_MALLOC_CONTEXT
> default on, but PB_ENABLE_MALLOC will remain default off.

ok

>    4. The alltypes style test cases are a bit of a maintenance
> burden. I'll probably merge custom_realloc to alltypes_pointer test
> case at some point. But I guess it is reasonable to have it separate
> for now when PB_ENABLE_MALLOC_CONTEXT remains a default-off feature.

ok

PetteriAimonen:
I agree, in your use case where allocator releases all allocations made in context, `pb_release()` is optional.

For general use, having stream passed to `pb_release()` makes most sense to me.
I think it could be made an optional parameter, and if the stream is `NULL` then `pb_release()` would internally do:

    pb_istream_t default = {NULL, NULL, 0, &pb_free_wrapper};
    if (!stream) stream = &default;

To provide a reasonable upgrade path to get clean API for 0.5 and also backward compatibility for 0.4 series, I suggest the following on header side:

    #ifdef PB_ENABLE_MALLOC_CONTEXT
    /* stream is optionally used for error messages and to set realloc function, if NULL then pb_free() is used */
    void pb_release(pb_istream_t *stream, const pb_msgdesc_t *fields, void *dest_struct);
    #else
    void pb_release(const pb_msgdesc_t *fields, void *dest_struct);
    #endif

pb_decode.h
```diff
@@ -43,6 +43,10 @@ struct pb_istream_s
 #ifndef PB_NO_ERRMSG
     const char *errmsg;
 #endif
+  #ifdef PB_ENABLE_MALLOC_CONTEXT
+  /* Note: we do not need a corresponding free() here, because we can (and do) call realloc with size==0 to free memory. */
```
Et7f3, line 47: meta data might be keeped so it won't completely free ?

PetteriAimonen: All definitions of realloc I've seen fully free the object. See e.g. https://pubs.opengroup.org/onlinepubs/009695299/functions/realloc.html *"If size is 0 and ptr is not a null pointer, the object pointed to is freed."*

PetteriAimonen: Looks like C17/C23 deprecate the zero size.

PR #790, opened 2022-06-30
remove trailing spaces

td-krzysiek:
Hello,

This is just a small commit to remove trailing spaces when enum or struct code is generated without comments.


td-krzysiek:
The commit was included in #791 so this pull request is no longer needed, thanks!

Issue #533, opened 2020-05-18
Proto file options to convert generated ENUM and #defines to UPPERCASE


GrantRolls:
Convert package_name and enum name to uppercase when enum members are generated, and the helper #defines that accompany the enums.

Currently it does the following (long_names off)

```
#example.proto
package my_package

enum my_enum {
    MY_ENUM_ZERO = 0,
    MY_ENUM_ONE = 1,
}
```

generates to
```
//example.pb.h
typedef enum _my_enum {
    my_package_MY_ENUM_ZERO,
    my_package_MY_ENUM_ONE,
}my_enum ;

// #defines
_my_package_my_enum_MIN
```

where it would be nicer/consistent with standard formatting best-practices as

```
//example.pb.h
typedef enum _my_enum {
    MY_PACKAGE_MY_ENUM_ZERO,
    MY_PACKAGE_MY_ENUM_ONE,
}my_enum ;

// #defines
_MY_PACKAGE_MY_ENUM_MIN
```

PR #791, opened 2022-06-30
C style (WIP)

td-krzysiek:
Hi,

I've started to work on a pull request for #199 / #533. I would like to check if this is heading into the right direction, do you have any comments? I added a simple example `examples/c_style` to show generated code.

Kind regards,
Krzysztof


PetteriAimonen:
In a brief look, the generator side code changes look mostly reasonable.
I would suggest renaming `StyleDefault` to `NamingStyle` and then have `NamingStyleC` inherit from that, overriding the methods.
In that way there would be a common base class which can be useful for future maintenance.

The automatic tests detect some errors which must be fixed.

Also this feature will need an automatic test under e.g. `tests/namingstyle`.
This could have a proto file that has all the different element types, and then a C file that references them. If the reference by C names compiles, the feature works.
E.g. `test/typename_mangling` test case has similar structure.

td-krzysiek:
Thinking of this a bit longer, cli argument probably makes sense. This feature should be enabled for the entire code generation process, rather than enabled / disabled for certain fields etc.

PetteriAimonen:
Command line argument does make sense with the way the pull request code is currently made.
A per-file or per-message option could make sense also, but that would be more difficult to implement.

td-krzysiek:
Hi @PetteriAimonen,

I think this is now ready for a review, I still might have missed some flags or combination of flags which will break the code but I tried to cover most use cases. If you have idea what other options I can try, I'm happy to add this to the test case.

I wasn't able to change `*_MSGTYPE` defines to capital letters, these are merged internally with struct name in `PB_BIND` macro, and I don't want to touch `pb.h` in this commit. Hopefully this is still fine.

Let me now if you have comments or suggestion.

PetteriAimonen:
Yeah, `_MSGTYPE` will have to remain as is, because it has to match the struct name. But they are mostly for internal use anyway.

I'll try to find time to review in more detail in the next few days.

rosek86:
Thank you!

PetteriAimonen:
Thanks, this is merged now!

I decided to remove the example, because it didn't add much compared to the existing examples, so I felt it was just unnecessary maintenance burden.

PR #799, opened 2022-07-23
[Bazel] Allows importing nanopb.proto when defining proto libs

gbalke:
I needed to add these before I was able to import `nanopb.proto`.

Tested here with the following code sections:
https://github.com/BetzDrive/bldc-controller/blob/f5d91c1aa8525789434d0f0c76391e85edfd2169/firmware/proto/BUILD.bazel#L3-L8
https://github.com/BetzDrive/bldc-controller/blob/f5d91c1aa8525789434d0f0c76391e85edfd2169/firmware/proto/messages.proto#L1-L4

@silvergasp not sure if you had a different way of doing this. Since you started bazelifying this, I'd appreciate your input!

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
 load("@rules_python//python:defs.bzl", "py_binary")
 load("@nanopb_pypi//:requirements.bzl", "requirement")
 load("@rules_proto_grpc//:defs.bzl", "proto_plugin")
@@ -37,8 +38,8 @@ copy_file(
 py_binary(
     name = "protoc-gen-nanopb",
     srcs = [
-        ":protoc-gen-nanopb.py",
         "generator/nanopb_generator.py",
+        ":protoc-gen-nanopb.py",
     ],
     deps = [
         requirement("grpcio-tools"),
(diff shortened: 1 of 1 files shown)


silvergasp:
Yeah, that looks about right. It wasn't something that I personally needed so I forgot to add it.

BUILD.bazel
```diff
@@ -37,8 +38,8 @@ copy_file(
 py_binary(
     name = "protoc-gen-nanopb",
     srcs = [
-        ":protoc-gen-nanopb.py",
         "generator/nanopb_generator.py",
+        ":protoc-gen-nanopb.py",
```
silvergasp, line 42: Was the reordering of srcs here your formatting change or buildifier? I might have forgotten to format the plugin in my previous PR, though I'm unsure. In any case the plugins *should* be autoformatted to prevent thrash.

gbalke: I think gazelle caught it for me. I didn't manually move it!

gbalke: Yeah. I just tried manually moving it above and then re-ran gazelle and it moved it down.

silvergasp: Fair enough. Well, that's my fault then. Best to leave it as is and we can put a CI check in there at some point to prevent new changes that aren't formatted properly.

gbalke: Just to be clear; you want me to revert this line change?

silvergasp: If it's formatted with buildifier, then don't revert it.

PetteriAimonen:
Ok, so I guess ready to merge. Thanks both :)

PR #803, opened 2022-08-05
Generate MSGTYPE definitions for UENUM fields

computer-whisperer:
This causes MSGTYPE macros to generate for UENUM fields so that the enum types can be inferred by macro expansions. 

Without this there seems to be no way to deduce the type name of an enum in a protobuf message field from the preprocessor.

This allows for macro constructions that automatically call the ENUMTYPE_name() functions generated by the enum_to_string feature.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1422,7 +1422,7 @@ class Message(ProtoElement):
             result += '#define %s_DEFAULT NULL\n' % Globals.naming_style.define_name(self.name)
 
         for field in sorted_fields:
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
I think it would be better to make this a separate ENUMTYPE define. The MSGTYPE macros are used in the PB_SUBMSG_INFO_x macros internally also, so it might be confusing to mix enums to the same list when they are accessed only by user code.

Also add ENUM to the list while you are it :)

PetteriAimonen:
Thanks!

Issue #805, opened 2022-08-19
[Bazel] cc_nanopb_proto_library() does not find .options file


randyzhg:
`cc_nanopb_proto_library` does not find the `.options` file for my proto (it lives in the same directory and has the same name as the `.proto` file). I've found that I can point it to the options file using an absolute path via `extra_protoc_args`, but it seems to only take an absolute path. If `cc_nanopb_proto_library` cannot find options files automatically, is there at least a way to pass options files via relative paths?

What I'm doing now:
```
cc_nanopb_proto_library(
    name = "myproto_nanopb",
    protos = [":myproto_proto"],
    extra_protoc_args = [
        "--nanopb_plugin_opt=-f/asolute/path/to/myproto.options",
    ],
)
```

PetteriAimonen:
@silvergasp Do you want to take a look?

silvergasp:
Yeah, I'm aware of this one as its something I would also like. I've been experimenting with this a bit. But haven't got anything working yet.

silvergasp:
CC @tpudlik

PetteriAimonen:
In the SCons rules I solved similar by having the rule take both the .proto and .options files as source files. But I guess because the proto_library rule is shared with other protobuf libraries, that doesn't work so nicely here.

silvergasp:
Yeah I think that it should be doable with a new version of those shared protobuf libraries. There is a new attribute [extra_protoc_files](https://rules-proto-grpc.com/en/latest/lang/c.html) that is included in the newer version. So I could see it working something like this.

```python
cc_nanopb_proto_library(
    name = "myproto_nanopb",
    protos = [":myproto_proto"],
    extra_protoc_args = [
        "--nanopb_plugin_opt=-f$(location myproto.options)",
    ],
    extra_protoc_files = ["myproto.options"],
)
```
I haven't tested this approach yet. So it might take a little bit of experimenting.

randyzhg:
I just tried using `extra_protoc_files` and it didn't seem to work. `extra_protoc_files` expects a Bazel target because passing the absolute path yielded this:
```
...
in element 0 of attribute 'extra_protoc_files' in 'cc_nanopb_proto_compile' rule:
invalid target name
...
target names may not start with '/'
```
[...]

PR #807, opened 2022-08-22
Remove extraneous single quotes around value of --library-include-format option

randyzhg:
Remove extraneous single quotes around value of `--library-include-format`
option for the Bazel nanopb_plugin.

This bug only seems to surface when passing additional options to
`cc_nanopb_proto_library`, via either `extra_protoc_args` or `options`.

Example error message at compile time:
```
error: expected identifier or '(' before '\x622e6822'
    6 | '#include"pb.h"'
      | ^~~~~~~~~~~~~~~~
```

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -20,6 +20,7 @@ isotes <isotes@gmail.com>
 Maxim Khitrov <max@mxcrypt.com>
 Yaniv Mordekhay <yanivmo@users.noreply.github.com>
 Ming Zhao <mzhao@luminatewireless.com>
+Randy Zhang <randyzhg95@gmail.com>
 Google, Inc.
 Tom Roeder <tmroeder@google.com>
 Piotr Sikora <piotrsikora@google.com>
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -49,7 +49,7 @@ py_binary(
 proto_plugin(
     name = "nanopb_plugin",
     options = [
-        "--library-include-format='#include\"%s\"'",
+        "--library-include-format=#include\"%s\"",
     ],
     outputs = [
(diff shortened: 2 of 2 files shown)


PetteriAimonen:
I wonder what is happening here, now the tests fail with the generated code being `#includepb.h`. At least there is a space missing, but also the quotes seem to randomly come and go.

@silvergasp, @gbalke Do either of you want to take a look?

gbalke:
I'll try to give it a pass this evening.

silvergasp:
Yeah, I had to play with this a bit when I wrote the rules. However I didn't test with the extra args attribute. The quotes are definitely still required, as the default bracket style quotes didn't seem to play well with bazels sandbox. There is definitely an issue here. Though I'm fairly certain that removing the quotes isn’t the solution.

randyzhg:
Perhaps I should just close this pull request and file an issue instead?

silvergasp:
Perhaps... Though I'm curious as to how this worked for you in the first place. Did you test this with a set of 'extra_protoc_args'?

randyzhg:
Yeah so this came up when I was applying a workaround for finding the options file: https://github.com/nanopb/nanopb/issues/805

Basically I had to do this as a pair to override nanopb's `--library-include-format`
```
    extra_protoc_args = [
        "--nanopb_plugin_opt=-f"+"/absolute/path/myproto.options",
        "--nanopb_plugin_opt=-L#include\"%s\"",
    ],
```

silvergasp:
Yeah ok. That certainly helps in zeroing in on the issue.

randyzhg:
Closing because I filed issue https://github.com/nanopb/nanopb/issues/810

PR #808, opened 2022-08-23
Fix errors associated with spaces in file paths

TCWilliamson:
Proposing to add " " around paths to fix issues with commands failing due to spaces in file names or directories.

e.g. 
```
'C:\Users\{USERNAME}\OneDrive' is not recognized as an internal or external command,
operable program or batch file.
```

for
`'C:\Users\{USERNAME}\OneDrive - COMANYNAME\'`

I don't believe this should cause issues on other OS?

### DIFF
--- a/generator/platformio_generator.py
+++ b/generator/platformio_generator.py
@@ -53,12 +53,12 @@ else:
         print(f"custom_nanopb_protos: {nanopb_protos}")
         exit(1)
 
-    protoc_generator = os.path.join(nanopb_root, 'generator', 'protoc')
+    protoc_generator = '"' + os.path.join(nanopb_root, 'generator', 'protoc') + '"'
 
     nanopb_options = ""
-    nanopb_options += f" --nanopb_out={generated_src_dir}"
+    nanopb_options += f' --nanopb_out="{generated_src_dir}"'
     for opt in nanopb_plugin_options:
-        nanopb_options += (" --nanopb_opt=" + opt)
+        nanopb_options += (f' --nanopb_opt="{opt}"')
 
     try:
         os.makedirs(generated_src_dir)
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
@hacker-cb Do you want to take a look?

TCWilliamson:
Hold on, found some issues with the options. Working to resolve it now.

TCWilliamson:
Ok, could do with some insight on this, the --nanopb_opt is causing issues. (Also, can't find any reference for that option even existing?)

PetteriAimonen:
Yeah, protoc's `--<plugin>_opt` is undocumented I think.

TCWilliamson:
> Yeah, protoc's `--<plugin>_opt` is undocumented I think.

Any chance I could get it clarified so I can fix the code up to work?
Also, is the Platformio version up to date / what is done to update it to latest? (or do I need to just copy the whole thing into private libraries?)

hacker-cb:
@TCWilliamson I think that just adding braces to some of vars here is not good solution.  

Better to rename your username in Windows without spaces :)

TCWilliamson:
> @TCWilliamson I think that just adding braces to some of vars here is not good solution.
> 
> Better to rename your username in Windows without spaces :)

It's not my username, its a default name generated for a OneDrive directory. Need I add windows uses several folders with spaces in the name. 

I don't think it's unreasonable that code in 2022 should be able to support spaces in directories & file names, given it's a feature of basically every operating system.

PetteriAimonen:
The script not working with spaces in path is certainly a bug and should be fixed.

hacker-cb:
@PetteriAimonen .
Ok, I will fix it and I will write additional test in CI.

But keep in mind, that some paltforms, like ESP32 IDF will fail to build in any case because of their internals.

PetteriAimonen:
Should be fixed now, with a slightly different approach.

PR #809, opened 2022-08-24
PlatformIO generator fix spaces in path

hacker-cb:
This is replacement for the https://github.com/nanopb/nanopb/pull/808.

But it is not ready, because it has hack in `nanopb_generator.py`.

Actually I did not find any suitable solution to pass path with string to the `nanopb_generator.py`

I tried 

```
 nanopb_options += f' --nanopb_opt=\'-I"{proto_include_dir}"\''
```

Resulting command is:

```sh
"/Users/pavel/projects/nanopb/generator/../generator/protoc"  
--nanopb_out="/Users/pavel/projects/nanopb/examples/platformio test/.pio/build/pio_with_options/nanopb/generated-src" 
--nanopb_opt="--error-on-unmatched" 
--nanopb_opt="--verbose" --proto_path="/Users/pavel/projects/nanopb/examples/platformio test/proto" 
--nanopb_opt='-I"/Users/pavel/projects/nanopb/examples/platformio test/proto"' pio_with_options.proto
```

But protoc passes internal quotes inside value, and I got next  `nanopb_generator.py`:

**args:**
```
['--error-on-unmatched', '--verbose', '-I"/Users/pavel/projects/nanopb/examples/platformio test/proto"']
```

**options.options_path:**
```
['"/Users/pavel/projects/nanopb/examples/platformio test/proto"']
```

So, it has quotes inside and It doesn not work.

@PetteriAimonen lets continue discuss here if you have any ideas to make it more nice.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -2503,6 +2503,12 @@ def main_plugin():
     import os.path
     options.options_path.append(os.path.dirname(request.file_to_generate[0]))
 
+    # FIXME: Hack for https://github.com/nanopb/nanopb/pull/808
+    new_options_path = []
+    for p in options.options_path:
+        new_options_path.append(p.strip('"\''))
+    options.options_path = new_options_path
+
     # Process any include files first, in order to have them
     # available as dependencies
     other_files = {}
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Should be fixed now, with a slightly different approach.

PR #814, opened 2022-09-20
Allow including inside of extern C

jussike:
I have production code in C and I'm using C++ test framework and that caused some linkage problems with nanopb. This patch fixed my problem and hopefully some others too.


PetteriAimonen:
Thanks, I didn't even know about `extern "C++"`.

PR #822, opened 2022-10-12
Options subfolder

kurddt:
This fix the following use case where both `.options` and `.proto` are located in a folder hierarchy that need to be preserved for the generated files. Some of the proto files could be imported using their path related to the proto folder 

proto
├── A
│       ├── A.proto
│       ├── A.options
├── B
│      ├── B.proto
│      ├── B.options
│      ├── A.proto
│      ├── A.options
app
├── CMakeLists.txt
├── build
│        ├── A
│        │       ├── A.pb.h
│        │       ├── A.pb.c
│        ├── B
│        │      ├── B.pb.h
│        │      ├── B.pb.c
│        │      ├── A.pb.h
│        │      ├── A.pb,c

The issues where: 
* It was quite hard to make sure that protoc takes the proto folder has project root and not one of the subdirectory
* When the correct folder was taken as project root the plugin could not find options files (`A/A.options`  `B/B.options`) because `proto/A/` and `proto/B/` where added to the option file but not `proto/`.


PetteriAimonen:
Is it possible that this could break builds for some of the existing users of the CMake rules?

kurddt:
The change about the discovery of the option files won't break anything.

The modification of the order of import path might be more problematic but I think it should be OK. 
The following import path have been pushed back:
* `GENERATOR_PATH`: This one does not contains any .proto, so not sure why it's needed by probably shouldn't be the project root
* `GENERATOR_CORE_DIR`: This one contains nanopb.proto so it should not be considered as the project root 
* `CMAKE_CURRENT_BINARY_DIR`: This does not contains any .proto, o not sure why it's needed by probably shouldn't be the project root
So no issue there.

When RELPATH is not set no logic is changed. And when set I don't see a correct use case where a user would like to have the generated import dir before it.

kurddt:
I marked it ready for review, if you remember some recent contributor of the `FindNanopb.cmake` file it might be good to have their input on that

PetteriAimonen:
Thanks!

I checked a few of the recent CMake contributors earlier, but they aren't particularly active on GitHub currently.
This seems straightforward enough that I just merged it directly.

The `protoc` include path order can be a complex things so I'll add a migration note just-in-case, even though I don't expect this to cause problems for anyone.

Issue #835, opened 2022-11-21
Add pb_*.c files into Conan package


chughes-pika:
Going over the `README.md` in your project, there's the following line
> Include *pb_encode.c*, *pb_decode.c*, and *pb_common.c* in your project.

The `conanfile.py` that exists creates the following package when built 
```sh
$ mkdir build
$ cd build
$ conan install ..
$ conan build ..
$ conan package ..
$ ls package
package
├── bin
│   ├── nanopb_generator.py
│   └── protoc-gen-nanopb
├── conaninfo.txt
├── conanmanifest.txt
├── include
│   ├── pb_common.h
│   ├── pb_decode.h
│   ├── pb_encode.h
│   └── pb.h
├── lib
│   ├── cmake
│   │   └── nanopb
│   │       ├── nanopb-config.cmake
│   │       ├── nanopb-config-version.cmake
│   │       ├── nanopb-targets.cmake
│   │       └── nanopb-targets-release.cmake
│   └── libprotobuf-nanopb.a
└── local
    └── lib
        └── python3.10
            └── dist-packages
                └── proto
                    ├── nanopb_pb2.py
                    ├── nanopb.proto
                    └── _utils.py
```

Since the associated `*.c` files are necessary for compilation, I suggest updating the `conanfile.py` to include these into the built package. e.g.

```sh
...
├── src
│   ├── pb_common.c
│   ├── pb_decode.c
│   └── pb_encode.c
```

PR #836, opened 2022-11-21
Add the basic source files into conan package

chughes-pika:
# About
This PR adds the source files for compiling into the Conan package.  A proposed solution to https://github.com/nanopb/nanopb/issues/835

# Tested
Ran 
```sh
mkdir build
cd build
conan install ..
conan build ..
conan package ..
```
The output on the package directory is as follows 
```sh
package/
├── bin
│   ├── nanopb_generator.py
│   └── protoc-gen-nanopb
├── conaninfo.txt
├── conanmanifest.txt
├── include
│   ├── pb_common.h
│   ├── pb_decode.h
│   ├── pb_encode.h
│   └── pb.h
├── lib
│   ├── cmake
│   │   └── nanopb
│   │       ├── nanopb-config.cmake
│   │       ├── nanopb-config-version.cmake
│   │       ├── nanopb-targets.cmake
│   │       └── nanopb-targets-release.cmake
│   └── libprotobuf-nanopb.a
├── local
│   └── lib
│       └── python3.10
│           └── dist-packages
│               └── proto
│                   ├── nanopb_pb2.py
│                   ├── nanopb.proto
│                   └── _utils.py
└── src  # <----------------------------- This is the new import to the package
    ├── pb_common.c
    ├── pb_decode.c
    └── pb_encode.c
```


PetteriAimonen:
@amatosov-rbx @Psy-Kai @a1lu Do any of you want to take a look?

I'm not sure about the change to `0.4.7-dev` in the `conanfile.py`.
The version is updated automatically by `tools/set_version.sh`, which explicitly omits the `-dev`.
Not sure if there is a purpose for that, with some other build tools the `-dev` suffix confuses them.
So I think either it should be left as `0.4.7`, or `tools/set_version.sh` should allow `-dev` also.

chughes-pika:
> @amatosov-rbx @Psy-Kai @a1lu Do any of you want to take a look?
> 
> I'm not sure about the change to `0.4.7-dev` in the `conanfile.py`. The version is updated automatically by `tools/set_version.sh`, which explicitly omits the `-dev`. Not sure if there is a purpose for that, with some other build tools the `-dev` suffix confuses them. So I think either it should be left as `0.4.7`, or `tools/set_version.sh` should allow `-dev` also.

I'll revert that change.  I hadn't realized there was automated incrementation

a1lu:
> @amatosov-rbx @Psy-Kai @a1lu Do any of you want to take a look?

The change looks good to me so far, though I could not test it yet.

@chughes-pika Could you please show in a minimal example how the new source files are linked against a project?

Psy-Kai:
No one questioning this? o.o

Why would you pack the source files? Conan will generate a static library which you can link. There is just no need to bundle the source files. If you are afraid of performance issues or something, tweaking compiler parameter would be the right place to start (e.g. compiling with lto or something).

It would be better to remove the `include source files in your project` part from the readme since this is a bad practice that many embedded developer follow. Compiling and linking a static library is just fine. A static library in fact is just a bundle of object files. So compiling all files in a huge project (which leads to multiple object files which get linked) or linking a static library (which contains just object files which get linked) should lead to the same binary.

PetteriAimonen:
I have no experience with Conan, so not sure how things are usually done with it.

One point of note is the compile-time options such as `PB_ENABLE_MALLOC`, which would have to match between static library compilation and compilation of user code. The classic way is to have some separate header generated by autoconf, but for simplicity nanopb doesn't currently have such.

Psy-Kai:
> One point of note is the compile-time options such as `PB_ENABLE_MALLOC`, which would have to match between static library compilation and compilation of user code. The classic way is to have some separate header generated by autoconf, but for simplicity nanopb doesn't currently have such.

With conan you would add an [`option`](https://docs.conan.io/en/latest/reference/conanfile/attributes.html#options) which handles something like this.
The modern way would be that the build system sets the define. So conan would configure cmake to define or not define `PB_ENABLE_MALLOC`.

chughes-pika:
At the moment, I'm working to decouple some legacy build which are using git submodules to pull nanopb into an ARM build.  Is the recommendation here that the `conanfile.py` should be updated to include ARM as a viable build target?  I haven't worked with managing toolchains via Conan and was worried it'd be too large of a change to propose.

Psy-Kai:
This recipe is (should be) target independent as is.

PetteriAimonen:
Great that it got resolved, and thanks @Psy-Kai for clearing it up - I wouldn't have caught this.

@chughes-pika If you have time, an `examples/conan_simple` example project could be useful for others in the future.

chughes-pika:
> 

Sure thing!  I'll clean it up a bit from our conversation here and try to get a new PR up in the next day or two.

PR #837, opened 2022-11-26
Adding a simple Conan+CMake example

chughes-pika:
This adds a simple example for how to use `nanopb` as a Conan import on a CMake project


PetteriAimonen:
Thanks!

PR #838, opened 2022-11-29
Move extern block to accommodate enum functions

BattleDash:
The `enum_to_string` function is declared outside of the `extern "C"` block, making it unusable in C++ code.


PetteriAimonen:
Thanks for the fix!
I added a regression test also.

PR #834, opened 2022-11-21
Escape Python path (solve issue with a space)

ivankravets:
Related issue https://github.com/platformio/platformio-core/issues/4467

PR #839, opened 2022-12-07
PlatfromIO: Fix to work with spaces in pyhton path.

hacker-cb:
Related:
* https://github.com/platformio/platformio-core/issues/4467
* https://github.com/nanopb/nanopb/pull/834

### DIFF
--- a/generator/platformio_generator.py
+++ b/generator/platformio_generator.py
@@ -2,32 +2,33 @@ import os
 import hashlib
 import pathlib
 import shlex
+import subprocess
+
 import SCons.Action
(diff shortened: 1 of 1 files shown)


hacker-cb:
@ivankravets @PetteriAimonen 

Workflow was failed.

`/usr/bin/python3: No module named  pip`

I don't know the reason, may be there is some different environment in `env.Execute` and `subprocess.run`?

On my MacOS all ok...

PetteriAimonen:
When you pass the arguments as a list, `-m` and `pip` should be two separate entries

PR #840, opened 2022-12-08
Improve PlatformIO generator to handle spaces in paths

valeros:
- This PR mainly fixes #834 by placing path arguments to the Nanopb generator in the CMD array separately to avoid possible issues if paths contain spaces
- Other minor improvements


PetteriAimonen:
I think this does not result in any functional difference, but it is cleaner so I'll merge it.

PR #843, opened 2022-12-21
pb.h: Fix build on MSVC

rpavlik:
Apparently MSVC lacks C-style static asserts, but it does have C++ style static asserts even in C code. Monado has used nanopb with this patch for some time and it seems to work fine.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -170,6 +170,9 @@ extern "C" {
 #    if defined(__ICCARM__)
        /* IAR has static_assert keyword but no _Static_assert */
 #      define PB_STATIC_ASSERT(COND,MSG) static_assert(COND,#MSG);
+#    elif defined(_MSC_VER)
+       /* MSVC has static_assert keyword but no _Static_assert */
+#      define PB_STATIC_ASSERT(COND,MSG) static_assert(COND,#MSG);
 #    elif defined(PB_C99_STATIC_ASSERT)
        /* Classic negative-size-array static assert mechanism */
 #      define PB_STATIC_ASSERT(COND,MSG) typedef char PB_STATIC_ASSERT_MSG(MSG, __LINE__, __COUNTER__)[(COND)?1:-1];


PetteriAimonen:
Can you elaborate on which MSVC version and with what compiler options?
The automatic CI builds run with Visual Studio 2019 and `/std:c11`, and haven't needed this option.

If this is specifically about C99 standard, it makes sense to make the `#ifdef` conditional on that.

Wallbraker:
All of them fail to build in Monado CI, `_Static_assert` isn't available on our CI at least.

Wallbraker:
https://gitlab.freedesktop.org/ryan.pavlik/monado/-/jobs/33803634#L242

Wallbraker:
Okay it seems `/std:c11` did the trick, or in our case setting `set(CMAKE_C_STANDARD 11)`.

PetteriAimonen:
Ok, it seems that MSVC [defaults to some C89/C99 hybrid](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170).

For general programming, it certainly makes sense to use standards from this century.
But I added a commit that should fix this in branch now, if it is easy for you, could you check if commit e57ec4a works without `/std:c11`?

rpavlik:
Yes, that does seem to work. yeah the C support in MSVC is not great and very weird by default.

PetteriAimonen:
Thanks for testing!

Wallbraker:
Thank you for the change, yeah agreed setting `c11` is probably not a big ask. We just hadn't done it, but doing it now. But probably is good to also support the default hybrid `C89/C99` as well.

PetteriAimonen:
@akashthakur76 reported in [commit comments](https://github.com/nanopb/nanopb/commit/ed530007aa04b31b32fc056e49df4a8fd1080da5#commitcomment-101889222) that they get `"expected declaration specifiers or '...' before numeric constant" in line 200: PB_STATIC_ASSERT(1, STATIC_ASSERT_IS_NOT_WORKING)` with MSVC in C89 mode after the latest change.

@akashthakur76 Can you tell what compiler settings you are using and which version of MSVC?

akashthakur76:
Well thanks Petteri. I'm able to fix the problem now. The comments given in pb.h helped in fixing.

PR #847, opened 2023-01-03
Ensure generator files are updated when cached

hansbinderup:
When bumping Nanopb, while having the previous version cached, the generator files are not correctly copied during build.

CMake copy_directory does not seem to copy the content if the directory does already exist.

By specifying that we want to copy the content of the directory we can avoid this issue.


hansbinderup:
> Thanks - could you also add a short comment to that line, to avoid the `/` getting accidentally dropped in the future? It's a quite subtle difference.

Thank you for the fast response :)

I've added a comment :+1:

PetteriAimonen:
Great, merged, thanks again :)

Issue #841, opened 2022-12-15
Cannot import 'TemporaryDirectory'


shahsahi:
I am currently trying to migrate from nanopb 0.4.5 to 0.4.7 and running into the following issue:

```
Traceback (most recent call last):
  File "...nanopb_generator.py", line 55, in <module>
    from proto import TemporaryDirectory
ImportError: cannot import name 'TemporaryDirectory' from 'proto' (unknown location)
```
For context, the way I am using nanopb is through a Conan package, and I am calling the generator directly with python (python nanopb_generator.py). The problem seems to be that the TemporaryDirectory import moved to `__init__.py`, which is not installed in CMakeLists.txt (the package install only has the following in the proto folder):

```
nanopb
└── lib
    └── site-packages
        └── proto
            ├── nanopb_pb2.py
            ├── nanopb.proto
            └── _utils.py
```

If I copy `__init.py__` here then the generator runs without issue. I can work around this but would it make sense for this file be installed alongside `_utils.py` in CMakeLists.txt?

PetteriAimonen:
I agree. I added the missing file to CMakeLists.txt now.
Somewhat related to #556.

It would be great to have an automatic test for cmake install or conan package.
Similar to bazel, platformio, ios_swift and pypi_publish workflows here:
https://github.com/nanopb/nanopb/tree/master/.github/workflows

There is already `examples/conan_dependency`.
I wonder if just adding a workflow that builds it according to the README would be enough?

shahsahi:
This fixes my issue, thanks! And yes seems like a workflow according to the README might make sense as well.

PetteriAimonen:
The main issue should be fixed in 0.4.8. I added separate issue #922 about adding a continuous integration test for the Conan build.

PR #848, opened 2023-01-04
Fix PYTHON_INSTDIR resolve in Docker containers

Crucio32000:
Hi everyone,
I've found an issue while creating a Docker container (using as a base image _crops/yocto:ubuntu-20.04-base_ ) meant to use nanopb code generator.
It would always resolve such variable into:
**/usr/local/local/lib/python3.10/dist-packages/proto/nanopb_pb2.py**

As you can see, the _/local/local_ should be resolved into _/local/_ only. To fix this, i've made a very small fix on this branch.

Out of curiosity, the current approach does work in the following myself-tested environments:
- Windows
- Yocto (both native and target)

NB: This is my first time opening a pull request on such project. Please be kind!
Thank you!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,7 +42,7 @@ endif()
 find_package(Python REQUIRED COMPONENTS Interpreter)
 execute_process(
     COMMAND ${Python_EXECUTABLE} -c
-        "import os.path, sys, sysconfig; print(os.path.relpath(sysconfig.get_path('purelib'), start=sys.prefix))"
+        "import os.path, sys, sysconfig; from pathlib import Path;  print(Path(sysconfig.get_path('purelib')).resolve())"
     OUTPUT_VARIABLE PYTHON_INSTDIR
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )


PetteriAimonen:
As far as I can tell, the current logic is:

* Get the Python library directory relative to Python's `sys.prefix`, put the relative path in `PYTHON_INSTDIR`
* Install Python files to `CMAKE_PREFIX_PATH/PYTHON_INSTDIR`

With this pull request, the logic would be:

* Get the absolute Python library directory, put it in `PYTHON_INSTDIR`
* Install Python files to this absolute path

The latter is certainly simpler for the system-wide install, but it would probably cause problems when installation is done to a different prefix (similar to issue #296). The issue #555 also contains useful information.

In your case, it seems you have:

* `sys.prefix = "/usr"`
* `sysconfig.get_path('purelib') = "/usr/local/lib/python3.10/dist-packages"`
* `CMAKE_PREFIX_PATH = "/usr/local"`

I think the correct solution for path resolving might be:

* Get the Python library directory relative to `CMAKE_PREFIX_PATH` if it is under it. Otherwise take it relative to `sys.prefix`. Put the relative path in `PYTHON_INSTDIR`.
* Install Python files to `CMAKE_PREFIX_PATH/PYTHON_INSTDIR`

@dflogeras What do you think about the CMake path resolving?

---

But that still just slams a package called `proto` in the Python lib directory, which is not very neat.
With the changes introduced in #460, we could install it properly as a package called `nanopb` instead, like `pip install nanopb` does. But that requires the `.../bin/protoc-gen-nanopb` to be generated as a wrapper script that calls `nanopb.generator.nanopb_generator.main_cli` - I'm not sure if it is pip or something else that generates that.

@theunkn0wn1 Do you happen to know what would be the best way to do the equivalent of `poetry build` and `pip install` in the CMake-based installation, while allowing to specify an installation prefix?

---

(As an aside, I'm still lacking automatic tests for CMake builds, as noted in #841 also. So we should work out something similar to `.github/workflows/platformio_tests.yml` to cover installs to home folder, `/usr` and to `/usr/local`. But this can be done separately.)

Crucio32000:
Hi @PetteriAimonen ,
Thank you for the prompt and very complete response, also tracking down the history that lead to current implementation.
So far, in my environment, i can tell that:
- Evaluating the `Python_INSTDIR` with this pull request (through _execute_command_ ) leads to `local/lib/python3.10/dist-packages` 
- Expanding variable `Python_SITELIB` expands into `/usr/lib/python3/dist-packages` .

I'm not so sure what its the best approach to solve this and work on every platform, but for sure you will catch any future regression if leveraging Github workflows.

As far as i understood:
- `protoc-gen-nanopb` is needed to let `protoc` enable the plugin. This, in any case, is referring to a python module located somewhere in the system (usually in a `{dist, site}-packages` directory)
- The python module is not correctly installed with CMake on every environment, that's **where the problem lies** currently.

Maybe we can take advantage of **PyInstaller** to generate a standalone binary ?
I can further investigate on my own once we decide a path to follow.

Thank you!

theunkn0wn1:
> Maybe we can take advantage of **PyInstaller** to generate a standalone binary ?


Py installer is no solution. 

1. It only works to produce windows bundles
2. Using it incurs significant technical debt, and has a history of causing difficult to debug issues. 
3. It is non portable and contradicts the standard for distributing python packages.

--------

> Do you happen to know what would be the best way to do the equivalent of poetry build and pip install in the CMake-based installation, while allowing to specify an installation prefix?

Iirc poetry does allow a build.py to be specified, enabling build scripts, however it is unstable and the Poetry devs show no interest in committing to formalizing it. (See https://github.com/python-poetry/poetry/issues/2740 )

PetteriAimonen:
I agree on pyinstaller, I'm using it for generating the binary packages but it doesn't make sense for CMake builds.

One potential option would be to leverage Python's zip support, so that `protoc-gen-nanopb` would be a zip file of source code files containing the proto folder also. But that's probably not very clean solution either.

Poetry's build.py support wouldn't help much anyway. I think at some point it may be necessary to reorganize the nanopb directory layout to be closer to what Python expects. Then, hopefully, we could easily copy a directory called `nanopb` to Python's lib folder and have `protoc-gen-nanopb` find it there. Currently poetry/pip/something generates a wrapper script for this, but I'm not sure what would be the proper way to replicate that in CMake.

PetteriAimonen:
There is now a potential solution here: #907.

PR #862, opened 2023-04-07
pb_decode.c: Fix possible OOB Vuln in 'pb_decode_inner'

Cezarus27:
Possible memory illegal accesses if `req_field_count = PB_MAX_REQUIRED_FIELDS`. In that case the right shift `(req_field_count >> 5)` should be 2 and it may want to access index 2 of `fields_seen.bitfield` which could cause crash.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -201,10 +201,10 @@ PB_STATIC_ASSERT(1, STATIC_ASSERT_IS_NOT_WORKING)
 
 /* Number of required fields to keep track of. */
 #ifndef PB_MAX_REQUIRED_FIELDS
-#define PB_MAX_REQUIRED_FIELDS 64
+#define PB_MAX_REQUIRED_FIELDS 64u
 #endif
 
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Hmm, I don't see the vulnerability. Do you have a code that actually causes OOB access?

In the case where req_field_count is 64, the loop on line 1132 goes through indexes 0 and 1, like it should.
The if on line 1141 never executes because `(64 & 31) != 0` is not true.

As for cases where the count is more than 64, they are covered by the compile-time check generated from https://github.com/nanopb/nanopb/blob/master/generator/nanopb_generator.py#L2130-L2134

It might make sense to have a check in the code in any case, but it might make sense to have it return error instead of silently ignoring.

PetteriAimonen:
Any more details on this?

PetteriAimonen:
Closing this off due to lack of details.

PR #867, opened 2023-04-21
Add ability to specify size of generated enums

jheaff1:
While the [`packed_enum`](https://github.com/nanopb/nanopb/blob/91fb301dabb83beb28f2ad266be6aaee7a1a4e00/generator/proto/nanopb.proto#L76) option exists, this does not work when using the MSVC compiler. 

This PR adds the ability to specify the size of the generated enums (this works for C++ only)

### DIFF
--- a/docs/migration.md
+++ b/docs/migration.md
@@ -9,6 +9,16 @@ document.
 Nanopb-0.4.7 (2022-xx-xx)
 -------------------------
 
+### Add int_size option to enum fields
+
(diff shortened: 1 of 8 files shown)


PetteriAimonen:
Looks reasonable.

To be ready for merge, this still needs:

- Test case. I think this could be added to `tests/enum_sizes`, where the existing C program would verify that it compiles for plain C, and a new C++ unit test would test for the size specification.
- Migration note in `docs/migration.md`, because if someone has previously specified `MyMessage.*  int_size=IS_8`, it would now affect new fields.

jheaff1:
Apologies for the delay, I have added unit tests and updated the migration doc

PetteriAimonen:
Thanks, I'll take a look.

PetteriAimonen:
Everything looks fine, merged now, thanks again!

PR #868, opened 2023-04-25
ref |> Modified The generated code uses utf-8 encoding by default.

xiaoyu-ebox:
Avoid garbled code generated when annotating proto in Chinese or other languages.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -2499,7 +2499,7 @@ def main_cli():
             if dirname and not os.path.exists(dirname):
                 os.makedirs(dirname)
 
-            with open(path, 'w') as f:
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Which Python version are you using? I think unicode should already be default for Python3.

PetteriAimonen:
Hmm, actually depends on locale  https://stackoverflow.com/questions/36303919/what-encoding-does-open-use-by-default  so makes sense to force utf-8.

PetteriAimonen:
Seems to work well in my tests on both Python2 and Python3, thanks!

xiaoyu-ebox:
> 

I am using python3, which should be related to the windows system native language.

xiaoyu-ebox:
Thank you very much for this project. It helps me. Thank you.

Issue #873, opened 2023-05-10
missing "call" in add_custom_command of FindNanopb.cmake for MSVC


philippe44:
I think the custom_command of "FindNanopb.cmake" are missing a "call" prefix that causes them to never return properly in MSVC. As a result, .pb.c and .pb.h are not built. I suggest adding something like that
```
add_custom_command(
      OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${FIL_PATH_REL}/${FIL_WE}.pb.c"
             "${CMAKE_CURRENT_BINARY_DIR}/${FIL_PATH_REL}/${FIL_WE}.pb.h"
      COMMAND ${CALL_PREFIX} ${PROTOBUF_PROTOC_EXECUTABLE}
      ARGS -I${GENERATOR_PATH} -I${GENERATOR_CORE_DIR}
           -I${CMAKE_CURRENT_BINARY_DIR} ${_nanopb_include_path}
           --plugin=protoc-gen-nanopb=${NANOPB_GENERATOR_PLUGIN}
           ${NANOPB_OPT_STRING}
           ${ABS_FIL}
      DEPENDS ${ABS_FIL} ${GENERATOR_CORE_PYTHON_SRC}
           ${ABS_OPT_FIL} ${NANOPB_DEPENDS}
      COMMENT "Running C++ protocol buffer compiler using nanopb plugin on ${FIL}"
      VERBATIM )
```
Where CALL_PREFIX is defined to "call" under MSVC and nothing otherwise. It needs to be done there as well
```
  foreach(FIL ${GENERATOR_CORE_SRC})
      get_filename_component(ABS_FIL ${FIL} ABSOLUTE)
      get_filename_component(FIL_WE ${FIL} NAME_WE)

      set(output "${GENERATOR_CORE_DIR}/${FIL_WE}_pb2.py")
      set(GENERATOR_CORE_PYTHON_SRC ${GENERATOR_CORE_PYTHON_SRC} ${output})
      add_custom_command(
        OUTPUT ${output}
        COMMAND ${CALL_PREFIX} ${PROTOBUF_PROTOC_EXECUTABLE}
        ARGS -I${GENERATOR_PATH}/proto
          --python_out=${GENERATOR_CORE_DIR} ${ABS_FIL}
        DEPENDS ${ABS_FIL}
        VERBATIM)
  endforeach()
```
I'm not submitting a PR because I'm sure the maintainers will have a better option

PetteriAimonen:
[...]

PR #874, opened 2023-05-10
add "call" under Windows upon generator call

philippe44:
Under Windows, when cmake generates .vxproj, the call to the custom command must be done using "call" otherwise there is no return and the file is not generated. See https://github.com/nanopb/nanopb/issues/873


PetteriAimonen:
What happens if `PROTOBUF_PROTOC_EXECUTABLE` is an actual `protoc.exe` instead of `protoc.bat`?

philippe44:
It works - you can use "call" for an exec or a batch

PetteriAimonen:
Ok, thanks for confirming. Looks good to me.

Issue #845, opened 2022-12-23
The Criterion project is broken after nanopb upgrade to 0.4.7


yurivict:
https://github.com/Snaipe/Criterion

```
===>  Building for criterion-2.4.1
[  1% 1/51] /usr/ports/devel/criterion/work/Criterion-2.4.1/src/protocol/gen-pb.py /usr/local/bin/protoc /usr/local/bin/protoc-gen-nanopb ../src/protocol/criterion.proto ../src/protocol/criterion.options /usr/ports/devel/criterion/work/Criterion-2.4.1/_build/src
FAILED: src/criterion.pb.c src/criterion.pb.h 
/usr/ports/devel/criterion/work/Criterion-2.4.1/src/protocol/gen-pb.py /usr/local/bin/protoc /usr/local/bin/protoc-gen-nanopb ../src/protocol/criterion.proto ../src/protocol/criterion.options /usr/ports/devel/criterion/work/Criterion-2.4.1/_build/src
/usr/local/bin/protoc --plugin=protoc-gen-nanopb=/usr/local/bin/protoc-gen-nanopb -I../src/protocol --nanopb_out=-f ../src/protocol/criterion.options:/usr/ports/devel/criterion/work/Criterion-2.4.1/_build/src ../src/protocol/criterion.proto
Traceback (most recent call last):
  File "/usr/local/bin/protoc-gen-nanopb", line 7, in <module>
    from nanopb_generator import *
  File "/usr/local/bin/nanopb_generator.py", line 55, in <module>
    from proto import TemporaryDirectory
ImportError: cannot import name 'TemporaryDirectory' from 'proto' (unknown location)
--nanopb_out: protoc-gen-nanopb: Plugin failed with status code 1.
ninja: build stopped: subcommand failed.
```

PetteriAimonen:
Seems like duplicate of #841. I'll release 0.4.8 in a few weeks to fix that, once any other breakages have been discovered.

paolostivanin:
Hello!
The fix for #841 helps only a bit. Now we get:

```
[    9s] Traceback (most recent call last):
[    9s]   File "/usr/bin/protoc-gen-nanopb", line 7, in <module>
[    9s]     from nanopb_generator import *
[    9s]   File "/usr/bin/nanopb_generator.py", line 53, in <module>
[    9s]     import proto
[    9s]   File "/usr/lib/python3.10/site-packages/proto/__init__.py", line 10, in <module>
[    9s]     import pkg_resources
[...]

PR #875, opened 2023-05-15
Add nanopb_PYTHON_INSTDIR_OVERRIDE to CMakeLists.txt

nicholas-stpierre-simplisafe:
Hello nanopb team,

I've been struggling to get local builds of nanopb integrated on various build systems. The pain point has been the nanopb `proto` python module installation location. To make this easier for myself and other nanopb users, I'd like to propose this change:

Add a `PATH` option to the cmake build named `nanopb_PYTHON_INSTDIR_OVERRIDE`. When set, cmake will use this path as the python install path and skip the `find_package(Python ...)` and `execute_process(COMMAND ${Python_EXECUTABLE} ...)` steps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,8 @@ option(nanopb_BUILD_RUNTIME "Build the headers and libraries needed at runtime" 
 option(nanopb_BUILD_GENERATOR "Build the protoc plugin for code generation" ON)
 option(nanopb_MSVC_STATIC_RUNTIME "Link static runtime libraries" ON)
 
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Yeah, the issue of `proto` being used as a Python system-wide package name in the CMake rules only came to my attention after the previous release. Some discussion in #845 and #848. I think the correct solution for that one would be to make the package name `nanopb` instead, similar to how `pip install nanopb` does it.

But as for Python module installation directory, it probably makes sense to be able to set it separately in any case.

nicholas-stpierre-simplisafe:
thanks very much for the quick response!

PR #888, opened 2023-08-11
Fix first path passed to protoc

twam:
Fix relative path passed to protoc when using RELPATH but NANOPB_GENERATE_CPP_APPEND_PATH not set. In this case the path provided by RELPATH should be passed to protoc first instead of the current source directory.

This was working before and broke with 8cc860c.

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -144,7 +144,7 @@ function(NANOPB_GENERATE_CPP SRCS HDRS)
       list(APPEND _nanopb_include_path "-I${ABS_PATH}")
     endforeach()
   else()
-    set(_nanopb_include_path "-I${CMAKE_CURRENT_SOURCE_DIR}")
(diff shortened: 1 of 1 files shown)


twam:
@kurddt can you have a look as you did the changes in 8cc860c?

PetteriAimonen:
It would be great to have some test cases for this, but I don't really know how each CMake option is supposed to be used.

twam:
I never wrote any tests for cmake stuff :(

kurddt:
@twam the line 
https://github.com/nanopb/nanopb/blob/8ef41e0ebd45daaf19459a011f67e66224b247cd/extra/FindNanopb.cmake#L147

should be changed to 
```c
list(APPEND _nanopb_include_path "-I${ABS_PATH}")
```
I currently don't have any way to test this so could you check it and make a PR if it works for you ?

twam:
Thanks for the quick check. That also works fine.

PetteriAimonen:
Thanks both!

Issue #890, opened 2023-08-15
ModuleNotFoundError: No module named 'nanopb_generator' using Bazel and Python 3.11


kersson:
**Steps to reproduce the issue**

1. Import `nanopb` into your bazel `WORKSPACE`
2. Use a Python 3.11 interpreter
3. Try to build a `cc_nanopb_proto_library` target

Sample files:
```
nanopb-bazel-test
├── BUILD.bazel
├── WORKSPACE
└── test_proto.proto
```

`WORKSPACE`:
```bazel
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

http_archive(
    name = "rules_python",
    sha256 = "0a8003b044294d7840ac7d9d73eef05d6ceb682d7516781a4ec62eeb34702578",
    strip_prefix = "rules_python-0.24.0",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.24.0/rules_python-0.24.0.tar.gz",
)

git_repository(
    name = "com_github_nanopb_nanopb",
    commit = "4ea09e8adcbd27e04f818117e88365e288ab8837",
    remote = "https://github.com/nanopb/nanopb.git",
)

load("@com_github_nanopb_nanopb//extra/bazel:nanopb_deps.bzl", "nanopb_deps")

nanopb_deps()

load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")

py_repositories()

python_register_toolchains(
    name = "python3_11",
    python_version = "3.11",
)

load("@python3_11//:defs.bzl", "interpreter")
load("@com_github_nanopb_nanopb//extra/bazel:python_deps.bzl", "nanopb_python_deps")

nanopb_python_deps(interpreter)

load("@com_github_nanopb_nanopb//extra/bazel:nanopb_workspace.bzl", "nanopb_workspace")

nanopb_workspace()
```

`BUILD.bazel`:
```bazel
load("@rules_proto//proto:defs.bzl", "proto_library")
load("@com_github_nanopb_nanopb//extra/bazel:nanopb_cc_proto_library.bzl", "cc_nanopb_proto_library")

proto_library(
    name = "test_proto",
    srcs = ["test_proto.proto"],
)

cc_nanopb_proto_library(
    name = "test_nanopb",
    protos = [":test_proto"],
)
```

`test_proto.proto`:
```proto
syntax = "proto3";

message Foo {
[...]

PR #891, opened 2023-08-15
Add missing source and data files to `protoc-gen-nanopb`

kersson:
This should fix https://github.com/nanopb/nanopb/issues/890 by adding all the required source and data files needed for the `protoc` generator.


PetteriAimonen:
Looks reasonable to me.

@silvergasp Do you want to take a look?

PR #892, opened 2023-08-16
Support stripping package names with a dependent proto file

### LINKED ISSUE #865: mangle_names not working on v0.4.7
There is a bug when file_a.proto includes file_b.proto which has a package definition.  Using mangle_names=M_STRIP_PACKAGE in both .proto files eliminates the enum and package names added to the front of the enum variables but file_a.pb.h still uses the long name in the _init_default initializer.  Since the _MIN version of the long name doesn't exist and isn't handled by the translation from long name to short name, the generated code will fail to compile on GCC.  

1. Generate the .c/.h files from the attached file_a.proto and file_b.proto.
2. Compile file_a.pb.h, file_a.pb.c, file_b.pb.h, file_b.ph.c using gcc.

**What happens?**

GCC will fail with an error due to _error_package_GenericErrorType_MIN not being defined.

file_b.pb.h has:
   #define GenericError_init_default                {_error_package_GenericErrorType_MIN, _error_package_GenericErrorSeverity_MIN}

file_a.pb.h has:
   #define _GenericErrorType_MIN GENERIC_ERROR_UNKNOWN

**What should happen?**

The code compiles.  Either file_a.pb.h needs to use the shortened names in the enum initializers (_init_default) or the long names need to be defined in the file_b.pb.h along with the long name translations. 

I changed the file extension so it could be posted here. 
[file_a.pb.h.txt](https://github.com/nanopb/nanopb/files/11223446/file_a.pb.h.txt)
[file_a.proto.txt](https://github.com/nanopb/nanopb/files/11223447/file_a.proto.txt)
[file_b.pb.c.txt](https://github.com/nanopb/nanopb/files/11223448/file_b.pb.c.txt)
[file_b.pb.h.txt](https://github.com/nanopb/nanopb/files/11223449/file_b.pb.h.txt)
[file_b.proto.txt](https://github.com/nanopb/nanopb/files/11223450/file_b.proto.txt)
[file_a.pb.c.txt](https://github.com/nanopb/nanopb/files/11223451/file_a.pb.c.txt)

joshstrohminger:
Fixes #865

When proto file _B_ defines `MessageB` containing a field with type `MessageA` that is defined in proto file _A_ which is using name mangling, stripping package names in this case, the `_init_zero` and `_init_default` `#defines` in file _B_ still use the long names from file _A_ and fail to compile. Similarly, the `_MIN`, `_MAX`, and `_ARRAYSIZE` `#defines` only use shorter identifiers.

This solution tries to follow the existing behavior of creating a `#define` mapping from long names to short names when using name mangling. It checks if the identifier is mangled, and populates the `MangleNames.reverse_name_mapping` dictionary with both the identifiers, ensuring both conform to the naming style. I included a test for this behavior.

This also includes two other bug fixes, but I'm happy to split them out into separate PRs if desired:
1. Update readme instruction for testing on Mac.
2. Ensure long enum names conform to the naming style when using short enum names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ end in an error, the test cases were successful.
 
 Note: Mac OS X by default aliases 'clang' as 'gcc', while not actually
 supporting the same command line options as gcc does. To run tests on
-Mac OS X, use: `scons CC=clang CXX=clang`. Same way can be used to run
+Mac OS X, use: `scons CC=clang CXX=clang++`. Same way can be used to run
 tests with different compilers on any platform.
 
 For embedded platforms, there is currently support for running the tests
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -512,7 +512,7 @@ class Enum(ProtoElement):
             # Define the long names always so that enum value references
             # from other files work properly.
             for i, x in enumerate(self.values):
-                result += '#define %s %s\n' % (self.value_longnames[i], x[0])
+                result += '#define %s %s\n' % (Globals.naming_style.define_name(self.value_longnames[i]), Globals.naming_style.enum_entry(x[0]))
 
         if self.options.enum_to_string:
             result += 'const char *%s(%s v);\n' % (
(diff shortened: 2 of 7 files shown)


PetteriAimonen:
Looks great in first look. I'll check some more and merge today or tomorrow.

Thanks!

PR #897, opened 2023-09-01
Adding integer size overrides for fixed-width types.

charlie-resilient:
This PR adds integer size overrides for protobuf's `fixed32`, `sfixed32`, `fixed64`, and `sfixed64`. This should allow us to generate protos that use the `fixed32` encoded type and `uint8_t` C type, as an example.

## Usage

For example, the following proto:

```proto
message Sensor {
  fixed32 some_register = 1;
  sfixed32 another_register = 2;
}
```

with `IS_8` and `IS_16` set in the options file:

```
Sensor.some_register  int_size:IS_8
Sensor.another_register  int_size:IS_16
```

should generate the C struct:

```c
typedef struct _resilient_PPG {
    uint8_t some_register;
    int16_t another_register;
} Sensor;
```

I understand that on the wire, these types will still require the full 4 or 8 bytes, since there's no representation of length in the fixed-length encoding and there are no fixed 8 or 16-bit integers in the wire format.

## Context

I'm building an embedded system where CPU time is precious. The `varint`-encoding is CPU-intensive when compared to encoding the `fixed32` wire type. This motivates the desire to use the `fixed32` type (and relatives). At the same time, a number of the types in the firmware are `uint8_t` or `uint16_t` (e.g. register values). I'd like for the generated protos in C to use these same types, mainly to save memory. 

I started implementing these changes, but I'm new to the behind-the-scene details of `nanopb`, so please let me know what else might need changing to make this happen! For example, I'm guessing I need to update the encoder, as well as some tests. Either way, thanks for this awesome library :)


PetteriAimonen:
The remaining changes needed are:

1. Add check for `field->data_size` to [pb_decode_fixed32](https://github.com/nanopb/nanopb/blob/master/pb_decode.c#L1369-L1389), similar to how `pb_dec_varint()` does it. I think this could be separated to helper functions `static bool pb_assign_uint(pb_istream_t *stream, const pb_field_iter_t *field, pb_uint64_t value)` and `static bool pb_assign_int(... pb_int64_t value)` to do the overflow checking. It would be best if the normal path for 32-bit values could be fast like it is now.
2. Similar changes on encoder side, with variable-size helper functions `pb_load_uint()` and `pb_load_int()` separated from the current `pb_enc_varint()` code.
3. Add to test case `tests/intsizes`.

charlie-resilient:
Thanks @PetteriAimonen! I'll take a stab at this over the next few days.

charlie-resilient:
Quick question: How do we know whether the `fixed32` / `fixed64` field is signed or unsigned? This matters when choosing which of the `pb_assign_int` or `pb_assign_uint` helper functions to use during overflow checking (per your recommendation).

At first glance, it looks like both the signed and unsigned versions use the same `LTYPE` ([link](https://github.com/nanopb/nanopb/blob/master/pb.h#L859)). Do I need to add two new `LTYPE`s for the signed versions? I'll look in more detail tomorrow. Thanks!

PetteriAimonen:
Good point.. I don't think there is a way to know currently. And adding new LTYPEs for a minor feature like this sounds like a bit much.

Overall, I'm not sure this feature makes sense for the general nanopb codebase if it results in a huge amount of changes and new code.

charlie-resilient:
Agreed. It would be a tiny feature. Not worth the added complexity. Thanks for the discussion! Closing the PR.

PR #898, opened 2023-09-07
Fix Command Generator Location of Simple Example 

walidamriou:
This pull request addresses the issue of the command generator location. 

The change involves updating the command used for code generation from `../../generator-bin/protoc --nanopb_out=. simple.proto` to ` ../../generator/protoc --nanopb_out=. simple.proto` .

### DIFF
--- a/examples/simple/README.txt
+++ b/examples/simple/README.txt
@@ -20,7 +20,7 @@ run it with the command: ./simple
 On other platforms, you first have to compile the protocol definition using
 the following command::
 
-  ../../generator-bin/protoc --nanopb_out=. simple.proto
+  ../../generator/protoc --nanopb_out=. simple.proto
 
 After that, add the following five files to your project and compile:
 


PetteriAimonen:
Which platform is this for? Considering the makefile works for Linux (and OS X, BSD etc. even though readme doesn't say that) directly.

On Windows I guess it will work, if there is a suitable Python installation with necessary packages. Otherwise the binary package would be easier on Windows.

walidamriou:
This update effect README.txt only, the path of protoc is ` ../../generator/protoc --nanopb_out=. simple.proto `not  `../../generator-bin/protoc --nanopb_out=. simple.proto` in **examples/simple/README.txt** 

Tested in Windows

PetteriAimonen:
In the Windows binary packages, there is `generator-bin/protoc` instead. But yeah, it's not clear which would be better in the README.

walidamriou:
It's perfectly fine now. This is what I observed when using the master branch directly. 

It's possible that if I use the Windows binary packages, the experience might be different.

PR #899, opened 2023-09-14
Fix submessages with different mangling rules

ripatel-fd:
nanopb does not correctly generate C code when importing files with different mangling rules.

e.g.

foo.proto

```proto
syntax = "proto3";
package foo;

import "nanopb.proto";
option (nanopb_fileopt).package = "foo1";

import "bar.proto";

message A {
    .bar.B b;
}
```

bar.proto

```proto
syntax = "proto3";
package foo;

import "nanopb.proto";
option (nanopb_fileopt).package = "bar1";

message B {}
```

Will result in generated code like `foo1_bar_B` instead of `bar1_B`.

This happens because
- the Protobuf file is parsed before importing dependencies
- the field names are mangled before importing dependencies 
- the mangling naively always uses the **current package** mangling rules, instead of the other package's rules for imported submessages

Fixes this by
- using "canonical mangling" (i.e. `s/\./_/g`) instead of custom mangling when submessage is detected coming from a different package
- fixing up the "canonical mangled name" to the real mangled name in the dependencies step

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -584,7 +584,7 @@ class Field(ProtoElement):
         self.callback_datatype = field_options.callback_datatype
         self.math_include_required = False
         self.sort_by_tag = field_options.sort_by_tag
-        
+
         if field_options.type == nanopb_pb2.FT_INLINE:
             # Before nanopb-0.3.8, fixed length bytes arrays were specified
             # by setting type to FT_INLINE. But to handle pointer typed fields,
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
This needs a test case also. Related to #892 but I think the issue is separate?

PetteriAimonen:
Any updates on this? I can help along with the test case if you are busy.

PetteriAimonen:
No problem, I'll see about it at some point.

PetteriAimonen:
Thanks, merged via command line in commit 8163cd494b2b2 and added test case in 7d790f9eb

PR #904, opened 2023-10-11
Add support for .options files in bazel rules

tobithiel:
Making a PR out of the suggestion in https://github.com/nanopb/nanopb/issues/805#issuecomment-1338272685 to support `.options` files in the provided bazel rules.

I added a bazel test to test the new support (failing on first commit, passing on second commit).

PR #906, opened 2023-10-12
Add support for nanopb options files without patching rules_proto_grpc

mark64:
This builds on #904 to add support for multiple nanopb options files and does so in a way that works without patching rules_proto_grpc for #805.

This should also let you specify options files that are built by bazel.


PetteriAimonen:
@tobithiel How does it look to you?

tobithiel:
Looks great. Awesome to have a way without the manual patch.

PR #913, opened 2023-10-30
Add requirements.txt file

MaJerle:
This pull requests proposes python `requirements.txt` file with listed module dependencies.

A simple run of `pip install -r requirements.txt` will automatically load required libs.
It can be future updated with new libs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+protobuf
+grpcio-tools


PetteriAimonen:
Thanks.

It would be nice to somehow ensure that the protobuf package version is compatible with the grpcio package version.
By default `pip install protobuf grpcio-tools` sometimes decides to use system's already installed `protobuf-python` and proceeds to install the newest and incompatible `grpcio-tools`.
But there is probably nothing we can do about that on nanopb side, it should be solved in grpcio-tools package.

MaJerle:
What we can do is to specify fixed grpcio-tools version in the requirements file.

https://pip.pypa.io/en/stable/reference/requirements-file-format/

PR #916, opened 2023-11-05
pb_byte_t: Add option to be uint8_t

MaJerle:
This pull requests add option to force `pb_byte_t` to be an 8-bit type. With the current implementation, `uint_least8_t` isn't guaranteed to be as low as `8-bits`. Compiler will throw a warning, if you pass the bytes array as parameter to a function accepting `uint8_t*` type.

Furthermore, if `uint_least8_t` isn't 8-bits, alignment won't work properly in a function, accepting `uint8_t`.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -52,6 +52,10 @@
  * Normally it is automatically detected based on __BYTE_ORDER__ macro. */
 /* #define PB_LITTLE_ENDIAN_8BIT 1 */
 
+/* This can be defined to force pb_byte_t to be uint8_t type, otherwise
+ * it is defined as uint_least8_t which is not guaranteed to be as low as 8-bits type */
+/* #define PB_BYTE_UINT8 */
+
 /* Configure static assert mechanism. Instead of changing these, set your
  * compiler to C11 standard mode if possible. */
 /* #define PB_C99_STATIC_ASSERT 1 */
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
uint_least8_t will equal uint8_t if the compiler supports 8-bit types.

The pull request can be acceptable, but you'll need to fix the comment to be more specific on when this is actually useful.

Though I think instead of a separate option, it might make sense to make it dependent on `#if defined(UINT8_MAX)` or similar, if the purpose is just to silence type mismatch warnings on some compilers. Would require some reading of the C standard or references to see what is valid.

MaJerle:
> uint_least8_t will equal uint8_t if the compiler supports 8-bit types.
> 
> The pull request can be acceptable, but you'll need to fix the comment to be more specific on when this is actually useful.
> 
> Though I think instead of a separate option, it might make sense to make it dependent on `#if defined(UINT8_MAX)` or similar, if the purpose is just to silence type mismatch warnings on some compilers. Would require some reading of the C standard or references to see what is valid.

Yes your point about UINT8_MAX is better.

Standard says that least types must provide at at least desired value, but can be larger than this.

PetteriAimonen:
I think I'll add something like this after 0.4.8 is released.

I feel like there is some risk that it may break with some special compiler setup (like when processor doesn't support 8-bit types), so I don't want to merge it right before a release.

PetteriAimonen:
This is now merged in commit f79aadbe with some adaptations.

PR #925, opened 2023-11-21
library proporties proposal for Arduino Library registry

hpsaturn:
## Overview

The idea with this PR is try to have a better experience for Arduino users, trying to have the library in the Library Manager of Arduino IDE. Also to resolve some issues with other libraries that using Nanopb like a dependency in this ecosystem.

## Tasks

- [x] library.properties proposal (extracted from the PlatformIO manifest, library.json)
- [ ] fix link symbolic issue with `arduino-lint` tool (lookup for a solution)

## Issues

The implementation of Arduino IDE registry is very bad in comparation with PlatformIO registry, for instance Arduino Library Registry doesn't have support for link symbolic. If you pass the `arduino-lint` we have:

```bash
Linting library in /home/avp/pio/nanopb
ERROR: Symlink(s) found. Symlinks cause difficulties for Windows users. These block addition to the Arduino Library
```

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -1,0 +1,9 @@
+name=Nanopb
+version=0.4.8
+author=Petteri Aimonen <jpa@npb.mail.kapsi.fi>
+maintainer=Antonio Vanegas <hpsaturn@gmail.com>
+sentence=protocol buffers, protobuf, google
+paragraph=Nanopb is a plain-C implementation of Google's Protocol Buffers data format. It is targeted at 32 bit microcontrollers, but is also fit for other embedded systems with tight (<10 kB ROM, <1 kB RAM) memory constraints.
+category=Communication
+url=https://github.com/nanopb/nanopb.git
+architectures=*


PetteriAimonen:
This will also need:

* A github workflow for testing build using e.g. arduino-cli.
* Version number setting in `tools/set_version.sh`

I assume the symlink issue is caused by the `spm_headers/nanopb` folder?
@paulb777 Is there any way to avoid those symlinks? Could the swift build rules possibly create the symlinks when spm is building?

hpsaturn:
Thanks for your fast feedback. I added the workflow and the version changer, but please check this last one, I did test and works.

PetteriAimonen:
Those symlinks are necessary for Swift builds, so as-is this is not acceptable. Unfortunately it seems that Arduino does not currently support excluding files either (https://github.com/arduino/Arduino/issues/9787).

I'd also prefer for the Arduino workflow to actually build something, because just running lint will not catch most problems.

hpsaturn:
Well, but remember the idea original of this pull request: A simple manifest for the Arduino Library Registry. Well, maybe usando `arduino-cli` we could run a complete build test.. or on other hand I have this library, [ESPNow Joystick Library](https://github.com/hpsaturn/espnow-joystick), using Nanopb with PlatformIO and Arduino Framework, we could have a test running one of my examples..

PetteriAimonen:
Doesn't have to be anything special, e.g. just build examples/simple using arduino-cli. It already verifies that all dependencies are found correctly.

hpsaturn:
make sense, but for that we need first the manifest, I guess. I'm not expert in Arduino CLI, I hate it, they have a very bad implementation compared with PlatformIO. Whatever, I think step to step, first we would have this PR, after that we would need the next tasks:

- [ ] make a PR to Arduino Registry to include the library (requirement: the manifest)
- [ ] make a PR here for change the arduino-lint with a special flag after Arduino approval the first task
- [ ] make a PR with a example using the library and arduino-cli

hpsaturn:
(I forgot say that I did an improvement to fix the Swift workflow)

hpsaturn:
Hello, could I get some feedback here? The main goal and its validation are complete. Is there anything specific holding up this PR?

PetteriAimonen:
I'm sorry, I have probably missed some notification.

I don't think the Swift fix is suitable, as it does the copying only in the Github workflow. That wouldn't work for any actual users of the Swift build in their own applications.

hpsaturn:
Don't worries. well, what can we do? maybe change the current symbolic links to a copy of these files? or add this step in the documentation for iOS users, and remove this directory, how to is now in this PR or proposal? Thanks for your comments.

PetteriAimonen:
I do not know and currently I don't have time to find out.

In my opinion the Arduino registry is defective in not allowing symlinks in the repository. I would be fine with getting rid of them, but that requires someone who knows how the Swift package manager works. I have never used it, and @paulb777 might now know or have time either.

It's a bit difficult maintaining the contributed build rules when I don't have knowledge of the platforms, so there is little I can do at the moment to move this forwards. In maintenance I put backwards compatibility first, so if SPM and Arduino platforms are incompatible, the SPM stays because it was there first.

PR #927, opened 2023-12-08
[bazel] Add support for generating python protos that include nanopb, use bazel-built protoc

mark64:
This makes 2 changes:
1. Adds the ability for us to generate python protobuf code from protos that include `nanopb.proto`.
2. Modifies the `PATH` for the protoc plugin such that it invokes the bazel-built `protoc` instead of the system installed `protoc` in `invoke_protoc`.


PetteriAimonen:
Hmm, why is the PATH wrapper needed? When invoked as a protoc plugin, it shouldn't need to call protoc. At least unless nanopb_pb2.py needs to be rebuilt? Or is that exactly the purpose?

mark64:
> Hmm, why is the PATH wrapper needed? When invoked as a protoc plugin, it shouldn't need to call protoc. At least unless nanopb_pb2.py needs to be rebuilt? Or is that exactly the purpose?

Yeah it looks like that's it, I hadn't dug into exactly why before.

If I make `nanopb_py_proto` a dependency of `protoc-gen-nanopb`, then it should theoretically not need `protoc` itself. However, I then run into an issue with `load_nanopb_pb2` in `generator/proto/__init__.py` because it assumes the protobuf will be located in the same directory, but bazel places generated code in a separate `.runfiles` directory and modifies `sys.path` to make the imports look the same.

Should the `load_nanopb_pb2` code change instead to somehow detect that it's being run inside a bazel environment and not do any rebuilds or manual file checks, instead just relying on a `try` import of `nanopb_pb2`?

mark64:
What I mean is that this patch on top of this PR will work, but I'm not sure how you'd prefer to fix `__init__.py` since I can see that check being useful outside a bazel environment

```
diff --git a/BUILD.bazel b/BUILD.bazel
index 69ac618..dfcb609 100644
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -31,7 +31,7 @@ cc_library(
 
 copy_file(
     name = "protoc-gen-nanopb.py",
-    src = "generator/protoc-gen-nanopb-bazel",
+    src = "generator/protoc-gen-nanopb",
     out = "generator/protoc-gen-nanopb.py",
     allow_symlink = True,
 )
@@ -45,13 +45,12 @@ py_binary(
     ],
     data = glob([
         "generator/**/*.proto",
-    ]) + [
-        "@com_google_protobuf//:protoc",
-    ],
+    ]),
     imports = [
         "generator",
     ],
     deps = [
+        ":nanopb_py_proto",
         requirement("grpcio-tools"),
     ],
 )
diff --git a/generator/proto/__init__.py b/generator/proto/__init__.py
index cfab910..3aaacc4 100644
--- a/generator/proto/__init__.py
+++ b/generator/proto/__init__.py
@@ -72,27 +72,23 @@ def load_nanopb_pb2():
     tmpdir = os.getenv("NANOPB_PB2_TEMP_DIR")
     temporary_only = (tmpdir is not None)
     dirname = os.path.dirname(__file__)
+    # XXX this path is correct when run in bazel
     protosrc = os.path.join(dirname, "nanopb.proto")
+    # XXX but this path is not, so nanopb thinks this file doesn't exist.
     protodst = os.path.join(dirname, "nanopb_pb2.py")
     proto_ok = False
 
     if tmpdir is not None and not os.path.isdir(tmpdir):
         tmpdir = None # Use system-wide temp dir
 
-    if os.path.isfile(protosrc):
-        src_date = os.path.getmtime(protosrc)
-        if not os.path.isfile(protodst) or os.path.getmtime(protodst) < src_date:
-            # Outdated, rebuild
-            proto_ok = False
-        else:
-            try:
-                from . import nanopb_pb2 as nanopb_pb2_mod
-                proto_ok = True
-            except Exception as e:
-                sys.stderr.write("Failed to import nanopb_pb2.py: " + str(e) + "\n"
-                                "Will automatically attempt to rebuild this.\n"
-                                "Verify that python-protobuf and protoc versions match.\n")
-                print_versions()
+    try:
+        from . import nanopb_pb2 as nanopb_pb2_mod
+        proto_ok = True
+    except Exception as e:
+        sys.stderr.write("Failed to import nanopb_pb2.py: " + str(e) + "\n"
+                        "Will automatically attempt to rebuild this.\n"
+                        "Verify that python-protobuf and protoc versions match.\n")
+        print_versions()
 
     # Try to rebuild into generator/proto directory
     if not proto_ok and not temporary_only:
diff --git a/generator/protoc-gen-nanopb-bazel b/generator/protoc-gen-nanopb-bazel
deleted file mode 100755
index fc726ee..0000000
--- a/generator/protoc-gen-nanopb-bazel
+++ /dev/null
@@ -1,14 +0,0 @@
-#!/usr/bin/env python3
-# This file is used to invoke nanopb_generator.py as a plugin
-# to protoc on Linux and other *nix-style systems.
-# Use it like this:
-# protoc --plugin=protoc-gen-nanopb=..../protoc-gen-nanopb --nanopb_out=dir foo.proto
-import os
-
-from nanopb_generator import *
-
-if __name__ == '__main__':
-    os.environ['PATH'] = 'external/com_google_protobuf:' + os.environ.get(
-        'PATH', '')
-    # Assume we are running as a plugin under protoc.
-    main_plugin()
```

PetteriAimonen:
Yeah, having to build nanopb_pb2.py has always been a mess.

If importing `nanopb_pb2` works, maybe it could use `nanopb_pb2.__file__` instead of the hardcoded path for the up-to-date check? Though if it is out of date, it would then need to unload and reload the generated file, which I think doesn't work with Google's Python protobuf library.

Maybe there is some way in Python to look up where it would import `nanopb_pb2` from, so it could get the path before importing?

mark64:
Checking for the file turns out to not be sufficient, the source file could have a more recent timestamp but bazel won't necessary rebuild the file if the SHA of the contents didn't change.

Also, the import path is slightly different under bazel

mark64:
@PetteriAimonen Friendly ping on this, would you like @tobithiel to review?

PetteriAimonen:
Sorry for the delay.

Looks good to me, except I think the environment variable could be useful for other builds than just bazel.
I made a small change in https://github.com/nanopb/nanopb/commit/0fb645b92dac00e5df574df599a50f0f8b53b61f , could you pull from there and test if it works correctly for you?

Would it be easy to add some kind of test that would be run by the existing .github/workflows/bazel.yml that would test the nanopb.proto import? If it seems complex, I don't think it is strictly necessary.

mark64:
Tested fine on our repo. Hopefully this works for testing in CI

mark64:
The existing CI tests cover this now that we're depending on the nanopb py proto in bazel, see https://github.com/nanopb/nanopb/actions/runs/7663874563/job/20887356939

PetteriAimonen:
Thanks a lot!

PR #931, opened 2024-01-05
Add has_msgid() and msgid() methods to C++ message descriptors

festlv:
The existing C++ message descriptors are useful for use in templated message packing/unpacking code.
This PR adds has_msgid() and msgid() methods for the MessageDescriptor<> class.


festlv:
I gave this a bit more thought, and have my doubts on the following:
* if int32_t is the right type for message id?
* if the fields should be methods or just static const membeer variables?

PetteriAimonen:
I think the correct datatype would be `uint32_t`, considering [nanopb.proto definition](https://github.com/nanopb/nanopb/blob/master/generator/proto/nanopb.proto#L85).

Regarding function vs. static const variable, I'm not sure if it makes any difference. Function seems fine to me.

PR #933, opened 2024-01-11
Add message sizes to C++ wrappers.

### LINKED ISSUE #932: Add encoded size to C++ output
We have a message queue where we can push messages of different types to be funneled further by a task. The message handling is a lot of repetition based on the data type

```Cpp
case SendDisconnect:
    PublishMessageFromBuffer<Disconnect, Disconnect_size>(disconnectTopic, Disconnect_fields);
    break;
```

This is both tedious to read and error prone if you specify the wrong size (`Connect_size` vs `Disconnect_size`). 
Right now we're using the C interface and are considering switching to the C++ helpers. If the helpers also included the maximum size those calls could be made easier to read and harder to get wrong:

```Cpp
case SendDisconnect:
    PublishMessageFromBuffer<Disconnect>(disconnectTopic);
    break;
```

denravonska:
This adds a max_size field referring to the C #define, indicating the maximum encoded size of that message type. If the size cannot be determined, for example due to non-bounded repeated fields, the constant is omitted.

If anyone knows how to test for absence of types in C++ I'm all ears :)

This closes #932

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1525,7 +1525,7 @@ class Message(ProtoElement):
 
         size_define = "%s_size" % (self.name)
         if size_define in local_defines:
-            result += '    static PB_INLINE_CONSTEXPR const pb_size_t max_size = %s;\n' % (size_define)
+            result += '    static PB_INLINE_CONSTEXPR const pb_size_t size = %s;\n' % (size_define)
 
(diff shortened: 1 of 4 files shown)


denravonska:
There, I think I got it covered with fixed- and non-fixed repeated fields.

PetteriAimonen:
Thanks!

Issue #707, opened 2021-09-30
Add support to define options using precompiler symbols


labishrestha2:
Hi @PetteriAimonen! Thank you so much for creating this wonderful project! I've found it very useful and have leveraged it on several of my personal and professional projects. 

In one of my projects, I'd like to use the same protocol definition on several devices, but I'd like to allocate message/field storage space on a per-product basis. To achieve this, I'd like to add a way to define the message/field options using predefined symbols e.g.:

**MyMessages.proto**
```
message MyMessage {
      required string ID = 1;
}
```

**MyMessages.options**
```
MyMessage.ID                              max_size_symbol : "MY_MESSAGE_ID_SIZE"
```

Which would go on to generate:
```
typedef struct _MyMessage {
    char ID[MY_MESSAGE_ID_SIZE];
/* @@protoc_insertion_point(struct:MyMessage) */
} MyMessage;
```
and the message size as:
```
#define MyMessage_size                       (1 + MY_MESSAGE_ID_SIZE) // or whatever the correct size is
```

With this in place, I can define MY_MESSAGE_ID_SIZE outside of the generated folders on  per-project basis.

I wanted to open a issue report here to capture the feature and will create a new branch where I'll start adding such a feature. Please kindly let me know if there's already some way to achieve this. Thank you!

PetteriAimonen:
An alternate way is to use a separate `.options` file. Share the `.proto` between projects, but have individual `.options` in each project.

Regarding your suggestion, one risk is that users will define different values of `MY_MESSAGE_ID_SIZE` in different compilation units. But I guess as long as there is no default value, most users will figure out that it needs to be defined globally.

I'm nowadays extremely short on time that I can spend on nanopb, and the WIP patch seems quite extensive compared to the fairly rare usecase. Unfortunately I can't promise that I'll be able to review it in any reasonable timeframe.

labishrestha2:
Thank you for your consideration @PetteriAimonen! The separate `.options` way seems like a reasonable approach for now.
I'll keep tinkering with the code and leave the PR open in case anyone else would be interested to look into it - but feel free to remove it as you see fit. Regards!

pdgendt:
Hi @labishrestha2 & @PetteriAimonen 

I bumped into this issue, as this is just something I want! I use and maintain the `nanopb` module for the [zephyr rtos](https://docs.zephyrproject.org/latest/samples/modules/nanopb/README.html).

[...]

PR #936, opened 2024-01-29
Add cmake helper to automatically configure options.in files

pdgendt:
Add a cmake helper to allow auto-configuring options files as an alternative for #707


PetteriAimonen:
Thanks!

pdgendt:
> Thanks!

Thanks for merging! What's the ETA on a release/tag?

PetteriAimonen:
Probably a couple of months, I don't think there has been much development since last one so far.

PR #940, opened 2024-02-16
Add C++ wrappers to generated structs

zachdeibert:


PetteriAimonen:
I'm not immediately convinced that this is a good fit for nanopb as a project. It adds a bunch of extra code paths and thus maintenance load for just some syntactic sugar.

Also what is the idea behind putting all the data into structs and then calling ` AllTypes::encode_req_uint64(stream2, alltypes.req_uint64) ` etc. manually? Isn't that the whole purpose of `pb_encode()` that it goes through the fields automatically?

In general, I'm not sure how far nanopb should go towards C++, as it is still primarily a C library. There is e.g. Embedded Proto for a pure C++ library.

zachdeibert:
My goal with `encode_req_uint64`, etc. was to allow manual encoding (I added a C++ wrapper for every C method exposed by the library).  Would just `encode()` and `decode()` be better (drop operators, field iterators, `release()`, `get_encoded_size()`, manual encoding)?

PetteriAimonen:
I really don't like the idea of maintaining such a set of wrapper functions.

Maybe this is something that would fit the [nanopb_cpp](https://github.com/nanopb/nanopb_cpp) side project (which seems somewhat abandoned)?

I've just seen so many people come by, suggest some C++ wrappers and because I don't have interest in maintaining them, they end up as poorly functioning code.

PR #942, opened 2024-02-24
Add support for enabling features via CMake components

zachdeibert:


PetteriAimonen:
Hmm, does the test case actually test `cpp descriptors`? Instead of including the `simple.c`, it might be more useful to have something similar to https://github.com/nanopb/nanopb/blob/master/tests/cxx_descriptor/message_descriptor.cc , though no need to check as fully.

I think a basic file like:

      int main()
      {
          return MessageDescriptor<SimpleMessage>::size;
      }

would be enough to make sure the descriptors got generated, just by compiling it.

PR #946, opened 2024-03-01
Support zephyr module

pdgendt:
This PR allows using `nanopb` as a [zephyr](https://docs.zephyrproject.org/latest/services/serialization/nanopb.html) module directly, instead of maintaining/merging a specific downstream copy.


PetteriAimonen:
Thanks!

PR #947, opened 2024-03-12
Store all field descriptor parts in program memory on AVR to save SRAM

simon-wh:
We were having some issues with our builds targeting AVR based MCUs as we only have 8KB SRAM available and the `*_msg` and `*_submsg_info` constants were filling up our SRAM a decent amount. So I did some digging and expanded the PROGMEM support in nanopb to extend to the `*_msg` and `*_submsg_info` parts instead of just `_field_info`.

This has resulted in very significant savings to our builds. Averaging a 11% reduction in static SRAM allocations.

In my testing, this appears to work correctly in our usage of nanopb. However, we do not make any use of extension fields and the code for that is a bit trickier. I did some digging and I hope that it should also be working normally as well.

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -150,9 +150,20 @@ extern "C" {
 #include <avr/pgmspace.h>
 #define PB_PROGMEM             PROGMEM
 #define PB_PROGMEM_READU32(x)  pgm_read_dword(&x)
+#if defined(PB_FIELD_32BIT)
+#define PB_PROGMEM_READSIZE(x) PB_PROGMEM_READU32(x)
+#else
+#define PB_PROGMEM_READSIZE(x) pgm_read_word(&x)
+#endif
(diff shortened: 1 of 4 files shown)


PetteriAimonen:
Hmm, I would really want to keep pb_encode.c and pb_decode.c free of this AVR PROGMEM stuff. It is error prone and hampers reading the code.

simon-wh:
I can understand that, the AVR PROGMEM stuff is overall quite tricky to understand and definitely hurts the readability. It also makes maintenance a bit trickier as you potentially have to take it into account when making changes.

It's somewhat up to you at the end of the day, I'm not opposed to us just using our own patches on top for our builds if you don't want to take this into the mainline code

simon-wh:
FYI, I just rebased this onto the `maintenance_0.4` for our own purposes, so that's why there are some conflicts incase anyone is wondering

PR #950, opened 2024-03-15
Migrate to bzlmod

Sayter99:
This PR
* updates bazel deps
  * rules_python -> 0.31.0
  * rules_proto -> 6.0.0-rc2
  * protobuf -> 23.1
  * rules_proto_grpc -> 4.6.0
* migrate to bzlmod from workspace, more [details](https://bazel.build/external/migration)
* update bazel build doc

This is a **break change** that users would need to change the usage of the bazel module instead of the old workspace instructions as below (doc updated too).

```py
bazel_dep(name = "nanopb")
git_override(
    module_name = "nanopb",
    remote = "https://github.com/nanopb/nanopb.git",
    commit = "<commit>",
)
```


PetteriAimonen:
@tobithiel @mark64 @silvergasp Does any of you have time to give this a look?

The need to change bazel module inclusion should be added to docs/migration.md.
It sounds simple enough change that it can be done in middle of 0.4.x series.

Sayter99:
> It sounds simple enough change that it can be done in middle of 0.4.x series.

@PetteriAimonen just to confirm, does it mean adding a section for bazel under `Nanopb-0.4.9 (2024-xx-xx)`?

PetteriAimonen:
@Sayter99 Place is correct. Some cleanup may be needed, but let's wait for the general comments first in case there are any larger issues. I have close to zero Basel knowledge myself.

silvergasp:
At first glance this looks good, although it should probably be accompanied with a submission to https://github.com/bazelbuild/bazel-central-registry, rather than just using git_override. I'm not entirely up to speed on what is best practice when it comes to bzlmod so it might take a little longer for me to review this one. If @armandomontanez or @tpudlik have time to look at this then they would probably be better reviewers.

MODULE.bazel
```diff
@@ -0,0 +1,32 @@
+module(name = "nanopb")
```
armandomontanez, line 1: Add a version here.

docs/migration.md
```diff
@@ -25,6 +25,35 @@ to restore old behavior.
...
+* upgrade bazel deps
+  * rules_python: 0.31.0
+  * rules_proto: 6.0.0-rc2
+  * protobuf: 23.1
+  * rules_proto_grpc: 4.6.0
+* replace WORKSPACE by MODULE.bazel (bzlmod)
+
+**Required actions:** Change the way to import nanopb module in your project
+```py
+bazel_dep(name = "nanopb")
```
armandomontanez, line 44: Add a version here.

docs/migration.md
```diff
@@ -25,6 +25,35 @@ to restore old behavior.
...
+git_override(
+    module_name = "nanopb",
+    remote = "https://github.com/nanopb/nanopb.git",
+    commit = "<commit>",
+)
+```
+noted that the name of the module has been changed to `nanopb`, to better fit the convention of bzlmod.
+If the old name `com_google_protobuf` is preferred, can add `repo_name` parameter to indicate the repo name.
+```py
+bazel_dep(name = "nanopb", repo_name="com_google_protobuf")
```
armandomontanez, line 54: Add a version here.

docs/bazel_build.md
```diff
@@ -6,35 +6,21 @@ into the build.
 ## Getting started
 Add the following to your WORKSPACE file.
 ``` py 
-# WORKSPACE
-git_repository(
-    name = "com_github_nanopb_nanopb",
-    remote = "https://github.com/nanopb/nanopb.git"
-    commit = "<TODO:Enter your desired commit>",
+# MODULE.bazel
+bazel_dep(name = "nanopb")
```
armandomontanez, line 10: Add a version here.

MODULE.bazel
```diff
@@ -0,0 +1,32 @@
...
+pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
+use_repo(pip, "nanopb_pypi")
+pip.parse(
+    hub_name = "nanopb_pypi",
+    requirements_lock = "@nanopb//:extra/requirements_lock.txt",
+    python_version = "3.10",
+)
+
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
```
armandomontanez, line 25: Brief warning: I don't think Bazel will treat `http_archive` repositories as bzlmod deps. I suspect in this case it's okay because rules_proto_grpc has a working WORKSPACE file, but I can't remember off the top of my head. Just something to keep an eye out for.

armandomontanez (commented):
I strongly recommend adding bzlmod support in parallel rather than directly replacing the old WORKSPACE system. Many projects don't yet work with bzlmod, and frankly the experience isn't well-trodden. Also, until `nanopb` is added to the BCR every module that depends on something that specifies a `git_override` for `nanopb` will be broken unless they themselves specify a `git_override`.

Also, any time you declare or reference a module, ensure that there's a version attached to it. I've left a few comments to highlight the spots where this matters. This isn't just good practice: sometimes Bazel chokes if a dependency fails to specify a version. `0.3.9.11-dev` might be appropriate until the version can be aligned with a release.

BUILD.bazel
```diff
@@ -73,13 +73,29 @@ proto_plugin(
     visibility = ["//visibility:public"],
 )
 
+proto_library(
```
mark64, line 76: I no longer work at Astranis, but we needed the BUILD file to use the google protobuf instead of a separate protobuf for python. Otherwise, you end up with duplicate modules defining the same messages and it won't work at runtime. Please revert this unless it's necessary

MODULE.bazel
```diff
@@ -0,0 +1,32 @@
+module(name = "nanopb")
+
+bazel_dep(name = "bazel_skylib", version = "1.5.0")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_python", version = "0.31.0")
+bazel_dep(name = "rules_proto", version = "6.0.0-rc2")
```
mark64, line 6: Is the -rc2 version because the current release doesn't support bazel modules yet? I'd suggest a comment or holding off on merging the PR until released

.gitignore
```diff
@@ -29,5 +29,6 @@ examples/using_union_messages/encode
 generator/nanopb_pb2.pyc
 !generator-bin/**/*
 bazel-*
+MODULE.bazel.lock
```
mark64, line 32: I normally commit this so that people that checkout the repo can reproduce the exact same dependencies as me, if necessary.

docs/migration.md
```diff
@@ -25,6 +25,35 @@ to restore old behavior.
...
+```py
+bazel_dep(name = "nanopb")
+git_override(
+    module_name = "nanopb",
+    remote = "https://github.com/nanopb/nanopb.git",
+    commit = "<commit>",
+)
+```
+noted that the name of the module has been changed to `nanopb`, to better fit the convention of bzlmod.
+If the old name `com_google_protobuf` is preferred, can add `repo_name` parameter to indicate the repo name.
```
mark64, line 52: Isn't the old name `com_github_nanopb_nanopb`?

docs/migration.md
mark64, line 54: update the name

mark64 (changes requested):
For such a key library in the embedded ecosystem, I second the request to keep WORKSPACE support around. Bazel is great and I'd hate to ruin people's experience by forcing them to migrate to bzlmod to get the latest nanopb. I'd remove WORKSPACE support from nanopb a few months after bazel actually removes WORKSPACE support

mark64:
By the way thank you very much for this! The bzlmod transition is quite a lot of work for the general ecosystem but I really appreciate you submitting this improvement to an important repo!

armandomontanez:
+1 adding bzlmod support was on my todo list, thanks for jumping on it. 😁

.gitignore
Sayter99: good caught, I did it simply because I rarely see a repo committing this, for example, https://github.com/protocolbuffers/protobuf, https://github.com/rules-proto-grpc/rules_proto_grpc, etc.

As far as I know this will be generated while doing bazel build, is there any reason people cannot reproduce the exact same deps without the lock file? I feel that in that case, the problem is in the MODULE.bazel file not because not committing the lock file.

BUILD.bazel
Sayter99: interestingly, I found there are two `py_proto_library` commonly used. One is from `protobuf` repo, the other one is from `rules_python`. For the one in `protobuf`, I have to build python proto separately from source instead of using built "descriptor_proto" target. But the one from `rules_python` can do it. I switched to use it and make it much cleaner here.

MODULE.bazel
Sayter99: done

MODULE.bazel
Sayter99: no specific reason, just picked whatever latest https://registry.bazel.build/modules/rules_proto.
there is an older version `5.3.0-21.7` from bcr, too, which I think it implies bzlmod support is up, rc2 might refer to some new features of protobuf repo. but anyways I am not quite familiar with the naming convention for that repo, if you have a preferred version, I can modify it.

MODULE.bazel
Sayter99: you're right, it is not considered a bazel_dep. I found one release from bcr, switched to use it

docs/bazel_build.md
Sayter99: done

docs/migration.md
Sayter99: done

docs/migration.md
Sayter99: mb, done

docs/migration.md
Sayter99: done

Sayter99: done

Sayter99:
@mark64 @armandomontanez I agree that WORKSPACE support is important to have. I have spent some time on it but there was an internal python issue I don't actually know how to fix; if anyone can give it a try that would be great.

.gitignore
armandomontanez: It's often ignored because there's currently some usability issues with it: https://github.com/bazelbuild/bazel/issues/20369

It's okay to check in (that's what it was designed for), but until Bazel changes the behavior there might be some annoying churn. Because the Bazel changes to nanopb are relatively infrequent, it's probably fine to check it in.

mark64: What I mean is that `MODULE.bazel` just specifies version numbers and only for direct dependencies, but the lock file will say exactly the hash of the code it pulled in and all the transitive dep versions as well.

BUILD.bazel
```diff
@@ -5,7 +5,7 @@ load("@rules_proto_grpc//:defs.bzl", "proto_plugin")
 load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
 load("//extra/bazel:nanopb_cc_proto_library.bzl", "cc_nanopb_proto_library")
 load("@rules_python//python/pip_install:requirements.bzl", "compile_pip_requirements")
-load("@com_google_protobuf//:protobuf.bzl", "py_proto_library")
+load("@rules_python//python:proto.bzl", "py_proto_library")
```
mark64, line 8: Reposting your old comment since it got marked outdated:

> interestingly, I found there are two py_proto_library commonly used. One is from protobuf repo, the other one is from rules_python. For the one in protobuf, I have to build python proto separately from source instead of using built "descriptor_proto" target. But the one from rules_python can do it. I switched to use it and make it much cleaner here.

I'm a little confused here, `py_proto_library` from `com_google_protobuf` will implicitly include `descriptor_proto` for you, you don't need to add it. The old code worked properly with bazel, can you post the error you were getting?

If you want to change this, @tobithiel can you confirm this still builds for Astranis?

MODULE.bazel
mark64: I would use the latest released versions that are not `-rc` (release candidate) or `-alpha` versions, please update those and confirm they still work

mark64 (changes requested):
> I agree that WORKSPACE support is important to have. I have spent some time on it but there was an internal python issue I don't actually know how to fix; if anyone can give it a try that would be great.

Happy to take a look, can you update the PR with your WORKSPACE file and show me the error logs you're getting + example code you're using?

BUILD.bazel
Sayter99:
```sh
ERROR: /home/sayter/inno_ws/nanopb/BUILD.bazel:93:17: ProtoCompile generator/proto/nanopb_pb2.py [for tool] failed: (Exit 1): protoc failed: error executing ProtoCompile command (from target //:nanopb_py_proto_genproto) bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/protobuf~/protoc '--python_out=bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/generator/proto' -Igenerator/proto generator/proto/nanopb.proto

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
google/protobuf/descriptor.proto: File not found.
nanopb.proto:9:1: Import "google/protobuf/descriptor.proto" was not found or had errors.
nanopb.proto:144:12: "google.protobuf.FieldDescriptorProto.Type" is not defined.
nanopb.proto:173:8: "google.protobuf.FileOptions" is not defined.
nanopb.proto:177:8: "google.protobuf.MessageOptions" is not defined.
nanopb.proto:181:8: "google.protobuf.EnumOptions" is not defined.
nanopb.proto:185:8: "google.protobuf.FieldOptions" is not defined.
ERROR: /home/sayter/inno_ws/nanopb/BUILD.bazel:93:17: ProtoCompile generator/proto/nanopb_pb2.py failed: (Exit 1): protoc failed: error executing ProtoCompile command (from target //:nanopb_py_proto_genproto) bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/protobuf~/protoc '--python_out=bazel-out/k8-fastbuild/bin/generator/proto' -Igenerator/proto generator/proto/nanopb.proto

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
google/protobuf/descriptor.proto: File not found.
nanopb.proto:9:1: Import "google/protobuf/descriptor.proto" was not found or had errors.
nanopb.proto:144:12: "google.protobuf.FieldDescriptorProto.Type" is not defined.
nanopb.proto:173:8: "google.protobuf.FileOptions" is not defined.
nanopb.proto:177:8: "google.protobuf.MessageOptions" is not defined.
nanopb.proto:181:8: "google.protobuf.EnumOptions" is not defined.
nanopb.proto:185:8: "google.protobuf.FieldOptions" is not defined.
```

Since it couldn't fine `descriptor.proto`, I think it does not include `descriptor_proto`. The root cause should be version changed to protobuf module. To overcome it, two ways I have tried, 1) making another target for descriptor_proto_py, or 2) switch to py_proto_library from rules_proto.

Personally I like the 2)'s usage more, however, after testing it more. I found that I have to change 

I tested more for the two and am realizing that my original attempt should be better. Since `generator/proto/__init__.py` needs to be changed accordingly because that `py_proto_library` doesn't have "include" attribute, and will raise an error as below

```sh
ERROR: /home/sayter/inno_ws/nanopb/tests/bazel_workspace_support/BUILD:9:24: Compiling protoc outputs for nanopb_plugin plugin on target @//:test_proto_nanopb_pb failed: (Exit 1): bash failed: error executing ProtoCompile command (from target //:test_proto_nanopb_pb) /bin/bash -c 'mkdir -p '\''bazel-out/k8-fastbuild/bin/_rpg_premerge_test_proto_nanopb_pb'\'' && bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/com_google_protobuf_protoc_linux_x86_64/protoc.exe "$@"' ... (remaining 7 arguments skipped)

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
Traceback (most recent call last):
  File "/tmp/bazel-working-directory/__main__/bazel-out/k8-opt-exec-ST-13d3ddad9198/bin/external/nanopb/protoc-gen-nanopb.runfiles/__main__/../nanopb/generator/protoc-gen-nanopb.py", line 7, in <module>
    from nanopb_generator import *
  File "/home/sayter/inno_ws/nanopb/generator/nanopb_generator.py", line 67, in <module>
    nanopb_pb2 = proto.load_nanopb_pb2()
  File "/home/sayter/inno_ws/nanopb/generator/proto/__init__.py", line 86, in load_nanopb_pb2
    import nanopb_pb2 as nanopb_pb2_mod
ModuleNotFoundError: No module named 'nanopb_pb2'
--nanopb_plugin_out: protoc-gen-nanopb_plugin: Plugin failed with status code 1.
Use --verbose_failures to see the command lines of failed build steps.
```

Last, if there is a way to test it, I think it would be good to have it in the CI.

MODULE.bazel
Sayter99: there is no stable version for `rules_proto_grpc`, I think the decision has to be made to use `http_archive` or `alpha`, any thoughts?

BUILD.bazel
Sayter99: w.r.t. the python issue, it is exactly https://github.com/bazelbuild/rules_python/issues/1543.

I tried to downgrade rules_python and it works now. I created another PR targeting this branch https://github.com/Sayter99/nanopb/pull/1/files ; feel free to take a look. In that branch I downgrade rules_proto_grpc too since it is the latest stable version.

Sayter99: tl;dr @mark64 
1. changes - https://github.com/Sayter99/nanopb/pull/1/files
2. downgrade rules_python and rules_proto_grpc makes workspace working
3. tried to test different py_proto_library more but let's wait @tobithiel if there's a better way to test it

BUILD.bazel
tpudlik: Yeah, the `rules_python` version is the future. The `protobuf` implementation was just a stop-gap and will eventually be removed.

BUILD.bazel
tpudlik: It would be great to eventually migrate to the latest version of `rules_python` and use their `py_proto_library`, but maybe that can be done in a separate change.

Re the `ModuleNotFoundError`: I would have hoped this can be fixed by setting the `import_prefix` and `strip_import_prefix` on the `proto_library` appropriately (so that no `includes` are necessary).

Sayter99: @tpudlik ty,

> Re the ModuleNotFoundError: I would have hoped this can be fixed by setting the import_prefix and strip_import_prefix on the proto_library appropriately (so that no includes are necessary).

I have tried
```python
proto_library(
    name = "nanopb_proto",
    srcs = [
        "generator/proto/nanopb.proto",
    ],
    strip_import_prefix = "generator/proto/",
    import_prefix = "generator/proto/",
    deps = ["@com_google_protobuf//:descriptor_proto"],
)

py_proto_library(
    name = "nanopb_py_proto",
    deps = [
        ":nanopb_proto",
    ],
)
```
but no luck. Any idea?

> It would be great to eventually migrate to the latest version of rules_python and use their py_proto_library, but maybe that can be done in a separate change.

Makes great sense to me, however, I was not able to make it work when using rules_python > 0.26.0. I got the same error as https://github.com/bazelbuild/rules_python/issues/1543
```sh
The following chain of repository dependencies lead to the missing definition.
 - @rules_python_internal
This could either mean you have to add the '@rules_python_internal' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: Error computing the main repository mapping: cycles detected during computation of main repo mapping
```

It might come from deps of nanopb, I will look into it when I got some time.

mark64 (changes requested):
It looks like there's either something broken in the dependencies here, or the instructions are missing something.

I made a [test repo here](https://github.com/mark64/nanopb_bzlmod_test) so you can see what I mean. When I do `bazel build //...`, I get this error. The second error is after I modified `MODULE.bazel` to pull in version 0.31.0 of `rules_python`, which is a different error:

``` bash
gatta:~/repos/nanopb_bzlmod_test(0) main $ bazel build //...
Starting local Bazel server and connecting to it...
ERROR: Traceback (most recent call last):
	File "/home/mark/.local/tmp/bazel/_bazel_mark/fb812c6a486e2aa14b75223c468afd05/external/rules_python~/python/extensions/pip.bzl", line 295, column 21, in _pip_impl
		fail((
Error in fail: Default python version '3.11' missing in pip hub 'nanopb_pypi': update your pip.parse() calls so that includes `python_version = "3.11"`
ERROR: Analysis of target '//:packet_nanopb_proto_pb' failed; build aborted: error evaluating module extension pip in @@rules_python~//python/extensions:pip.bzl
INFO: Elapsed time: 5.691s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
FAILED:
    Fetching repository @@protobuf~; starting
    Fetching https://github.com/protocolbuffers/protobuf/archive/refs/tags/v23.1.zip; 1.9 MiB (28.3%)
    Fetching repository @@bazel_tools~cc_configure_extension~local_config_cc; starting
    Fetching module extension pip in @@rules_python~//python/extensions:pip.bzl; starting
gatta:~/repos/nanopb_bzlmod_test(1) main $ vim MOD
MODULE.bazel       MODULE.bazel.lock
gatta:~/repos/nanopb_bzlmod_test(1) main $ vim MODULE.bazel
gatta:~/repos/nanopb_bzlmod_test(134) main $ bazel build //...
ERROR: /home/mark/.local/tmp/bazel/_bazel_mark/fb812c6a486e2aa14b75223c468afd05/external/rules_python~~pip~nanopb_pypi/grpcio_tools/BUILD.bazel:27:6: configurable attribute "actual" in @@rules_python~~pip~nanopb_pypi//grpcio_tools:pkg doesn't match this configuration: No matching wheel for current configuration's Python version.

The current build configuration's Python version doesn't match any of the Python
versions available for this wheel. This wheel supports the following Python versions:
    3.10

As matched by the `@rules_python~//python/config_settings:is_python_<version>`
configuration settings.

To determine the current configuration's Python version, run:
    `bazel config <config id>` (shown further below)
and look for
    rules_python~//python/config_settings:python_version

If the value is missing, then the "default" Python version is being used,
which has a "null" version value and will not match version constraints.


This instance of @@rules_python~~pip~nanopb_pypi//grpcio_tools:pkg has configuration identifier 102280f. To inspect its configuration, run: bazel config 102280f.

For more help, see https://bazel.build/docs/configurable-attributes#faq-select-choose-condition.

ERROR: Analysis of target '//:packet_nanopb_proto_pb' failed; build aborted: Analysis failed
INFO: Elapsed time: 2.361s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
ERROR: Build did NOT complete successfully
FAILED:
    Fetching repository @@rules_pkg~; starting
    Fetching repository @@rules_python~~python~python_3_11_x86_64-unknown-linux-gnu; starting
```

Sayter99:
@mark64 thank you for testing it. I have replied to your test repo by making a branch. please have a look when you got time. thanks!

mark64:
Thanks for taking a look. I took your branch and made a few tweaks here so that nanopb doesn't force a specific python version on people (otherwise users need to use the same python toolchain as nanopb). You can find that code here: https://github.com/mark64/nanopb/tree/bzlmod-testing. I updated my repo to use it and now it passes.

It works for bzlmod, but it does not work for WORKSPACE. I put a comment in the BUILD.bazel, in order for it to work nanopb needs some way to use an alternative BUILD.bazel import and register the python toolchain as shown here: https://github.com/bazelbuild/rules_python/blob/46f4c25e716a7002453e4b24d5257a0913d77482/docs/sphinx/getting-started.md#toolchain-registration

Perhaps the right way to handle this is by adding a `.patch` file to nanopb and having WORKSPACE users add that patch to their repo in the `http_archive` fetch in their WORKSPACE? I'm not sure we can keep the `BUILD.bazel` compatible with both unless we do something like that.

Otherwise, the PR is looking nice, nearly ready to approve.

Sayter99:
@mark64 would you mind making another PR targeting to my branch, it overall looks good to me, I am more than happy to merge it.

> It works for bzlmod, but it does not work for WORKSPACE.

It means not work for WORKSPACE if using the latest rules_python, right? I think right now if we make different rules_python version for bzlmod and WORKSPACE it might work, although it looks hacky..

Anyways, I don't have strong opinion on how to support WORKSPACE build, feel free to let me know if there's anything I can do to help.

With regarding to MacOS CI failure, I tested it on my M1 MBA (MacOS 14), it works fine, thus I tried upgrading CI macOS to 14 and it works, too. I don't really have any idea why macos11 failed, if there's no major objection I'd prefer to upgrade MacOS version in bazel CI, too.

mark64:
Made a PR for you: https://github.com/Sayter99/nanopb/pull/2

It won't work for WORKSPACE using either the new nor old rules_python version. The repo I import in `BUILD.bazel` doesn't have an equivalent in the WORKSPACE world, so anyone who pulls in `nanopb` via WORKSPACE will need to patch the repo first to change that path.

For MacOS, will let @PetteriAimonen comment but I believe the intent is that nanopb be workable on a minimum-support version of MacOS. Switching to MacOS 14 may not be okay if he still wants to support olding MacOS releases.

Sayter99:
@mark64 sorry for the late reply, I have reviewed your change and left one comment, thank you! :)

PetteriAimonen:
Does anyone have any more comments before I merge this?

BUILD.bazel
```diff
@@ -1,5 +1,8 @@
 load("@rules_proto//proto:defs.bzl", "proto_library")
-load("@rules_python//python:defs.bzl", "py_binary")
+
+# XXX in bzlmod this works, but in workspace we need
```
mark64, line 3: The XXX comment should be replaced with something like this
```suggestion
# NOTE: if you're still using WORKSPACE, you'll need to patch this file to use
# the following instead
```

MODULE.bazel
mark64: Looks like this is still an http archive, did you intend to replace this?

Sayter99:
> Looks like this is still an http archive, did you intend to replace this?

@mark64 for grpc, 4.6.0 is the latest stable version; although it doesn't have bzlmod support, I think we can stick with it for a while and upgrade the version (which supports bzlmod) when the stable version is out.

BUILD.bazel
Sayter99: Done

PetteriAimonen:
Merged through squashed commit f6187b05a

Thanks everyone!

PR #957, opened 2024-04-25
Include C source files in pip package

mreed348:
## Commit message

Include all nanopb C source files as part of the Pip package to make updating nanopb more convenient for pip users.

Before this commit, updating nanopb took two steps: update the nanopb pip package, then update the nanopb C source files. Now updating nanopb takes only one step (i.e., update the pip package), which avoids the possibility of users updating the generator and forgetting to update the source files.

The source files show up in the root of the package:

    (.venv) ~/.../extra/poetry$ python -m zipfile --list dist/nanopb-0.4.9.dev1465-py2.py3-none-any.whl
    File Name                                             Modified             Size
    nanopb/__init__.py                             1980-01-01 00:00:00            0
    nanopb/generator/__init__.py                   1980-01-01 00:00:00            0
    nanopb/generator/nanopb_generator              1980-01-01 00:00:00          235
    nanopb/generator/nanopb_generator.bat          1980-01-01 00:00:00          206
    nanopb/generator/nanopb_generator.py           1980-01-01 00:00:00       112055
    nanopb/generator/nanopb_generator.py2          1980-01-01 00:00:00          460
    nanopb/generator/platformio_generator.py       1980-01-01 00:00:00         5839
    nanopb/generator/proto/Makefile                1980-01-01 00:00:00          126
    nanopb/generator/proto/__init__.py             1980-01-01 00:00:00         4851
    nanopb/generator/proto/_utils.py               1980-01-01 00:00:00         2790
    nanopb/generator/proto/google/protobuf/descriptor.proto 1980-01-01 00:00:00        36277
    nanopb/generator/proto/nanopb.proto            1980-01-01 00:00:00         7139
    nanopb/generator/proto/nanopb_pb2.py           1980-01-01 00:00:00         4443
    nanopb/generator/protoc                        1980-01-01 00:00:00         1577
    nanopb/generator/protoc-gen-nanopb             1980-01-01 00:00:00          374
    nanopb/generator/protoc-gen-nanopb-py2         1980-01-01 00:00:00          554
    nanopb/generator/protoc-gen-nanopb.bat         1980-01-01 00:00:00          449
    nanopb/generator/protoc.bat                    1980-01-01 00:00:00          302
    pb.h                                           1980-01-01 00:00:00        44076
    pb_common.c                                    1980-01-01 00:00:00        12141
    pb_common.h                                    1980-01-01 00:00:00         1677
    pb_decode.c                                    1980-01-01 00:00:00        53759
    pb_decode.h                                    1980-01-01 00:00:00         7378
    pb_encode.c                                    1980-01-01 00:00:00        30339
    pb_encode.h                                    1980-01-01 00:00:00         6958
    nanopb-0.4.9.dev1465.dist-info/METADATA        1980-01-01 00:00:00         5848
    nanopb-0.4.9.dev1465.dist-info/WHEEL           1980-01-01 00:00:00           92
    nanopb-0.4.9.dev1465.dist-info/entry_points.txt 1980-01-01 00:00:00          143
    nanopb-0.4.9.dev1465.dist-info/RECORD          2016-01-01 00:00:00         2430

### DIFF
--- a/extra/poetry/poetry_build.sh
+++ b/extra/poetry/poetry_build.sh
@@ -11,7 +11,7 @@ cd build
 ln -s ../dist .
 
 mkdir nanopb
-tar xf tmp.tar README.md generator
+tar xf tmp.tar README.md generator *.c *.h
 mv generator nanopb/
 touch nanopb/__init__.py nanopb/generator/__init__.py
 make -C nanopb/generator/proto
--- a/extra/poetry/pyproject.toml
+++ b/extra/poetry/pyproject.toml
@@ -10,7 +10,7 @@ homepage = "https://jpa.kapsi.fi/nanopb/"
 documentation = "https://jpa.kapsi.fi/nanopb/docs/index.html"
 keywords = ["protobuf", "protoc"]
 classifiers = ["Topic :: Software Development :: Build Tools"]
-include = ["nanopb/**/*", "nanopb/__init__.py"]
+include = ["nanopb/**/*", "nanopb/__init__.py", "*.c", "*.h"]
 
 [tool.poetry.scripts]
 nanopb_generator = "nanopb.generator.nanopb_generator:main_cli"


extra/poetry/poetry_build.sh
```diff
@@ -11,7 +11,7 @@ cd build
 ln -s ../dist .
 
 mkdir nanopb
-tar xf tmp.tar README.md generator
+tar xf tmp.tar README.md generator *.c *.h
```
mreed348, line 14: should we move the C source files under their own directory inside the pip package, e.g., under `src`?

PetteriAimonen:
Hmm, I'm unsure whether this makes sense.
Are there precedents of Python packages including C source files that should be compiled by user?
Is there a reasonable way to include those files in a build, considering they'll end up somewhere inside Python's library folders?

Nanopb generator is designed to have good compatibility for generated files, so having mismatched generator and source code versions is not necessarily a problem. But I agree that most users will want to update them at the same time. However, I would think it would make more sense in that case to get the full nanopb package (either as tar or as git submodule), which also includes the generator.

PetteriAimonen:
@theunkn0wn1 If you have time, do you have any thoughts on this? Btw. thanks for the initial pypi support back in #460, it has reduced installation problems for many people.

extra/poetry/poetry_build.sh
theunkn0wn1: Yes, these should exist under the main nanopb package so we can use `importlib.resources` to access them.

theunkn0wn1:
> do you have any thoughts on this?

So my thoughts here are somewhat mixed.
either way, the user needs access to the nanopb sources in order to compile the code this package generates.

When i was using this project, it was as simple as adding the nanopb package from `platformio`. 

I am hesitant to commit generated files to VCS, but true source files i have no real objections to.
Bundling the library sources into the pip package _should_ be fine I think.

Emitting them into the generated directory, however, should be behind an opt-in flag IMHO.
I would rather compilers not get confused by the presence of duplicate header files.

Bottom line: I think the idea of shipping the nanopb sources as part of the pip wheel **as long as these assets are not themselves generated** is fine. they can essentially be treated as [resources](https://importlib-resources.readthedocs.io/en/latest/using.html#file-system-or-zip-file).

We should make emitting these embedde sources optional, to retain compatibility with alternate sources of these files.



-------------
> thanks for the initial pypi support back in https://github.com/nanopb/nanopb/pull/460, it has reduced installation problems for many people.

:heart: 

NP. I recall the install process before my PR was driving me crazy :rofl:

PetteriAimonen:
Ok, so seems like this would be a reasonable approach.

@mreed348 Can you check how `importlib.resources` would best work for this, and then add some kind of comment or example on how it would be used? I'm not exactly sure where this should go though, as currently there isn't any documentation specific to the Python package.

PR #958, opened 2024-05-06
Prefer `generator/protoc` in CMake

### LINKED ISSUE #953: Protobuf 26.0 removes RegisterExtension
https://github.com/protocolbuffers/protobuf/commit/4ebba684c791416a9131dd01d4c5175a2beb3566

```
nanopb/generator/nanopb_generator.py -q -s packed_struct:0 -H.n.h -S.n.c metricblob.proto
Failed to import generator/proto/nanopb_pb2.py:
Traceback (most recent call last):
  File "nanopb/generator/proto/__init__.py", line 102, in load_nanopb_pb2
    from . import nanopb_pb2 as nanopb_pb2_mod
  File "nanopb/generator/proto/nanopb_pb2.py", line 22, in <module>
    google_dot_protobuf_dot_descriptor__pb2.FileOptions.RegisterExtension(nanopb_fileopt)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: type object 'FileOptions' has no attribute 'RegisterExtension'

Traceback (most recent call last):
  File "nanopb/generator/nanopb_generator.py", line 66, in <module>
    nanopb_pb2 = proto.load_nanopb_pb2()
                 ^^^^^^^^^^^^^^^^^^^^^^^
  File "nanopb/generator/proto/__init__.py", line 128, in load_nanopb_pb2
    return nanopb_pb2_mod
           ^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'nanopb_pb2_mod' where it is not associated with a value
```

pdgendt:
When building with CMake we should prefer to use the `protoc` executable that uses grpcio-tools.

Fixes #953

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,8 @@ option(nanopb_MSVC_STATIC_RUNTIME "Link static runtime libraries" ON)
 
 set(nanopb_PYTHON_INSTDIR_OVERRIDE "" CACHE PATH "Override the default python installation directory with the given path")
 
-find_program(nanopb_PROTOC_PATH protoc HINTS generator-bin generator)
+find_program(nanopb_PROTOC_PATH protoc PATHS generator-bin generator NO_DEFAULT_PATH)
+find_program(nanopb_PROTOC_PATH protoc)
 if(NOT EXISTS ${nanopb_PROTOC_PATH})
     message(FATAL_ERROR "protoc compiler not found")
 endif()
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -407,7 +407,7 @@ add_library(nanopb STATIC EXCLUDE_FROM_ALL ${NANOPB_SRCS})
 target_compile_features(nanopb PUBLIC c_std_11)
 target_include_directories(nanopb PUBLIC ${NANOPB_INCLUDE_DIRS})
 
-# Find the protoc Executable
+# Find the local protoc Executable
 find_program(PROTOBUF_PROTOC_EXECUTABLE
     NAMES protoc
     DOC "The Google Protocol Buffers Compiler"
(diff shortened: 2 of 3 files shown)


pdgendt:
@PetteriAimonen Not sure why the binary builds are failing, do these depend on CMake?

PetteriAimonen:
Binary build failure is mostly unrelated, I'm in progress of figuring that out in `dev_binpkg` branch.

PetteriAimonen:
If you rebase on master now, the build failures should be fixed.

PetteriAimonen:
Could you elaborate on what the changes to proto/__init__.py do, and why?
This affects a lot of use cases and build systems, so changes there need to be made quite carefully.

PetteriAimonen:
Ok, this looks reasonable and I also think that the `generator/protoc` (which uses grpcio if available and falls back to system protoc) should be preferred.

I think as currently written, there would be no easy way to force use of system protoc? I'm not sure if there is any situation where that would be needed.

@twam Do you have time to take a short look? The CMake side looks quite straightforward to me, if I understand correctly it just searches `generator-bin` and `generator` before system path.

pdgendt:
> Could you elaborate on what the changes to `proto/__init__.py` do, and why? This affects a lot of use cases and build systems, so changes there need to be made quite carefully.
> 
> (Edit: ah, I see where the undefined variable problem was coming from now - it just was such a bunch of changes that it was difficult to see :)

I separated 2 commits, each fixing 1 issue.

> I think as currently written, there would be no easy way to force use of system protoc? I'm not sure if there is any situation where that would be needed.

I think it should be possible to set `PROTOBUF_PROTOC_EXECUTABLE` before calling `find_package`.

twam (approved):
Looks good to me.

PetteriAimonen:
Thanks!

Issue #962, opened 2024-05-15
Suggestion: Required fields for proto3


canton7:
Hi,

We've been using nanopb extensively on a fairly large embedded project, for Ethernet/BLE communications with a number of devices, for file storage, and for logging. It's performed really well, and we've benefited a lot from it. Thank you!

There's one thing (with two aspects) which has been a bit of a source of pain however, and I'd like to suggest a way to improve it.

Proto3 (as you know) removed support for `required`, meaning that all fields which have a `message` type are implicitly optional, and so `nanopb` generates a corresponding `has_xxx field`.

However, on our embedded system many of these fields are not considered optional: they need to be present, and the message should be rejected if they are not. This means that we need to write validation. However, this validation code is best written at the point that the message is deserialized (it's otherwise hard to pass the correct errors back up through the call chain from where the message is actually used, C being what it is, and quite often messages are stored and then actually used later on). That's fine, but it means that the code which checks the presence of a `has_xxx` field is far removed from the code which actually accesses that field, and so it's hard to keep the two in sync.

The other aspect is the reverse: the embedded code will always populate some fields, however it needs to remember to set the `has_xxx` field at every point. This adds significant noise, and is also easy to forget (which has led to a few bugs). We also end up with some confusion over whether a field is actually semantically required to be set by the embedded side or not.

Additionally, we have some very large arrays of protobuf types (we're using it to serialize assay scripts to control a diagnostic instrument, and those can consist of many thousands of steps, each one being a protobuf message), and the `has_xxx` fields use up a significant amount of unnecessary memory.

My suggestion is to add a custom option which works the same as `required` for proto2: fields with this option set will have no corresponding `has_xxx` field, will always be serialized, and nanopb will return an error if they are not present on deserialization.

I'm aware of the reasons that `required` fields were removed in proto3, and I generally agree with them. However, I don't think they apply here, as a "required" option will only apply to nanopb and not to other parts of the system:

* For deserialization, it's functionally the same as manually writing code which validates that the corresponding `has_xxx` field is `true`.
* For serialization, it only states that the embedded device expects to send the field. Other devices in the ecosystem will be unaware of the option, and will therefore treat the field as optional. The option can therefore be removed later without breaking those devices.

We could achieve this by moving back to proto2, however the proto3 changes around enums and packed arrays are desirable. And, using proto2's `required` of course has downsides.

[...]

PR #963, opened 2024-05-15
Add label_override field option

canton7:
This lets proto3 files define submessage fields which nanopb treats as required (even though proto3 doesn't support 'required'). This means that nanopb won't generate a has_xxx field, will reject messages where this field is missing, and will always serialize the field. However, other consumers of the proto file won't see the field as required, and are therefore safe to it being removed in future.

To do this, use:

    Submessage some_submessage = 1 [(nanopb).label_override = LABEL_REQUIRED];

This mechanism also lets users change whether nanopb treats the field as optional or repeated in both proto2 and proto3.

Fixes: #962

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -615,6 +615,9 @@ class Field(ProtoElement):
             self.default = desc.default_value
 
         # Check field rules, i.e. required/optional/repeated.
+        if field_options.HasField("label_override"):
+            desc.label = field_options.label_override
+
         can_be_static = True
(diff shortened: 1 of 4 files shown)


canton7:
I had a bit of free time, so put this together.

Please feel free to hold off / reject if you're not sure of the benefit, or tell me what I've done wrong!

Thanks.

PetteriAimonen:
Looks great, thanks!

canton7:
Fantastic, thank you!

PR #965, opened 2024-05-21
Add supported GCC attributes for IAR compiler

josepho0918:
The GCC attribute `warn_unused_result` is supported by the IAR compiler starting from version 9.40.1.


PetteriAimonen:
Thanks! At some point it might make sense to add this for clang too.

PR #968, opened 2024-05-29
Do not include system paths when finding generator script

pdgendt:
Similar to searching for the local `protoc`, we should use the local `nanopb_generator.py` script.

The issue was reported at https://github.com/zephyrproject-rtos/zephyr/issues/70065#issuecomment-2133261857

### DIFF
--- a/extra/FindNanopb.cmake
+++ b/extra/FindNanopb.cmake
@@ -443,6 +443,7 @@ find_path(NANOPB_GENERATOR_SOURCE_DIR
     DOC "nanopb generator source"
     PATHS
     ${NANOPB_SRC_ROOT_FOLDER}/generator
+    NO_DEFAULT_PATH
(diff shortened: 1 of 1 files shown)


pdgendt:
Not sure what the CI issue is:

```
[2848] Error loading Python lib '/home/runner/work/nanopb/nanopb/nanopb-0.4.4-268-gb078cd0-linux-x86/generator-bin/_internal/libpython3.8.so': dlopen: /home/runner/work/nanopb/nanopb/nanopb-0.4.4-268-gb078cd0-linux-x86/generator-bin/_internal/libpython3.8.so: cannot open shared object file: No such file or directory
```

PetteriAimonen:
Thanks. No CI errors visible to me, but in any case that sounds like unrelated problem.

Issue #988, opened 2024-07-18
feature request: enum validation function


jaskij:
In some scenarios, I find the need to check if a value is a valid enum member. It would be great if there was an option to generate that validation function.

I see it very, very, similar to the way `enum_to_string` currently works - a function which takes a value of the type, and returns `true` if it's a known value, and `false` where `_name()` returns `"unspecified"`.

PR #991, opened 2024-07-19
Enums: add a validation function

jaskij:
As I wrote in #988 , adding an enum validation function is a feature I'd find useful.

The one big question I have is the type of argument the function should take. Should it be the type of enum used, or should it be some integer type?

Marking as draft until I add tests.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -521,7 +521,7 @@ class Enum(ProtoElement):
                 Globals.naming_style.type_name(self.names))
 
         if self.options.enum_validate:
-            result += 'bool *%s(%s v);\n' % (
+            result += 'bool %s(%s v);\n' % (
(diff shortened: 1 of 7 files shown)


jaskij:
The basic stuff should be done.

One more question: I see `enum_to_string` tested in the `namingstyle` test. Should I add something with `enum_validate` too?

PetteriAimonen:
Yeah, I think it would be good to have a test for naming style too, the name mangling sometimes has surprising bugs.

Other than that, looks ready to merge.

jaskij:
This should cover it?

Also, obvious caveat: this only checks if the integer value is a valid member of the enum, so it can throw a curveball, like in the naming style this passed.

```c
my_enum2_valid(ME_ENUM1_ENTRY_FIRST) == true
```

PetteriAimonen:
Thanks!

Issue #961, opened 2024-05-13
Incorrect generated data_size when using (nanopb_enumopt).int_size 


mrjoboto:
Hi there,
I’ve been experimenting with using the `(nanopb_enumopt).int_size` option to control the size of the generated enum. However, I’ve been running into issues during encode/decode when I set the enum size to 1 byte. Specifically, it seems like the message descriptor that gets generated for the enum field has an incorrect data_size value that doesn’t respect the int_size option used. But the generated struct does respect the int_size option. This leads to the encode function accessing memory beyond the end of the message structure. I tried to put together a contained example that shows the problem:
```
syntax = "proto3";

import "nanopb.proto";
option (nanopb_fileopt).long_names = false;

enum Foo {
  option (nanopb_enumopt).int_size = IS_8;
  kBar = 0;
  kBaz = 1;
}

message EnumContainer {
  Foo foo = 1;
}
```
The generated structures are what I would expect:
```
/* Enum definitions */
typedef enum _Foo
#ifdef __cplusplus
 : uint8_t
#endif
{
    kBar = 0,
    kBaz = 1
} Foo;

/* Struct definitions */
typedef struct _EnumContainer {
    Foo foo;
} EnumContainer;
```

However, the generated message descriptor seems to result in an inaccurate data_size for the enum field:
```
EnumContainer message = EnumContainer_init_zero;
message.foo = Foo::kBaz;

pb_field_iter_t iter;
pb_field_iter_begin_const(&iter, EnumContainer_fields, (void*)&message);
printf("%lu, %hu\n", sizeof(message.foo), iter.data_size);
```

For me this prints “1,4” - the size of the field is 1 byte, but the data_size value is 4 bytes, and data_size is used during encoding to iterate over a void* pointing into the message structure, leading it to access unrelated memory.

Any advice you have to offer would be most appreciated! I'm using nanopb 0.4.8 and clang 15.0.0 targetting arm64-apple-darwin22.6.0.

PetteriAimonen:
Hmm, are you perhaps generating the protocol definitions into a `.pb.c` file, but compiling your user code as `.cc` file?
Can you try putting definitions to `.pb.cc` instead?

I think `#ifdef __cplusplus` should be removed from the generator. Now C23 supports `: uint8_t` in C mode, and in any case it is better to have compiler error than runtime error.

mrjoboto:
Thanks for the prompt reply!

[...]

PR #992, opened 2024-07-19
generator: remove ifdef __cplusplus on enum sizes

jaskij:
While this does not solve #961 completely, it will make the compiler error when sized enums are supported.

I experimented somewhat with adding a preprocessor check, but it seems that GCC extends enum sizes to older C standards, since the enum_sizes test passes without it and fails to compile when I add that.

Yup, seems like GCC extends this extension to all versions of C.

This will absolutely break some builds, but at least it will be broken builds instead of faulty software.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -439,9 +439,7 @@ class Enum(ProtoElement):
         # Override the enum size if user wants to use smaller integers
         if (FieldD.TYPE_ENUM, self.options.int_size) in datatypes:
             self.ctype, self.pbtype, self.enc_size, self.data_item_size = datatypes[(FieldD.TYPE_ENUM, self.options.int_size)]
-            result += '\n#ifdef __cplusplus\n'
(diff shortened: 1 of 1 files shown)


jaskij:
Looking around: this is supported for all C standards since GCC 13. ARM GNU Toolchain 13 was released last year.

I am not entirely sure how to modify test scripts and/or CI to actually have the CI pass.

PetteriAimonen:
I wonder if it is the `-ansi` that makes GCC not like it in the test case.
The compiler options could be modified for that test case like is done in `anonymous_oneof/SConscript`

I wonder whether MSVC supports this.

jaskij:
> I wonder if it is the -ansi that makes GCC not like it in the test case.

That, and the compiler version. Best I can tell, this was added in GCC 13, possibly 12, and your CI runs on Ubuntu 20.04. Unless you're setting up a non-default compiler, that means GCC 9.

PetteriAimonen:
I think this is obsoleted by b34a649c524

Issue #887, opened 2023-08-08
DeprecationWarning pkg_resources is deprecated as an API


fermentedfly:
Hi,

I recently switched to Python 3.11 and now get Deprecation warnings for `nanopb_generator.py:22` -> `import pkg_resources`

Regards,
Manuel

PetteriAimonen:
This was originally included for PyInstaller, but appears to be no longer required.

PetteriAimonen:
Fix published in nanopb-0.4.8.

idan-ben-ami:
I get the same warning using any of these:

generator/proto/__init__.py:import pkg_resources
generator/proto/__init__.py:        _builtin_proto_include = pkg_resources.resource_filename('grpc_tools', '_proto')
generator/proto/_utils.py:            import pkg_resources
generator/proto/_utils.py:            paths.append(pkg_resources.resource_filename('grpc_tools', '_proto'))

It is required when importing nanopb_pb2.py from the generated python script (myproto_pb2.py).

lubosz:
Because of this `nanopb_generator` fails to run on Python 3.12:

```
$ nanopb_generator
Traceback (most recent call last):
  File "/home/bmonkey/.local/bin/nanopb_generator", line 5, in <module>
    from nanopb.generator.nanopb_generator import main_cli
  File "/home/bmonkey/.local/share/pipx/venvs/nanopb/lib/python3.12/site-packages/nanopb/generator/nanopb_generator.py", line 57, in <module>
    from . import proto
  File "/home/bmonkey/.local/share/pipx/venvs/nanopb/lib/python3.12/site-packages/nanopb/generator/proto/__init__.py", line 10, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

jaskij:
To test for this, I modified `binary_packages.yml` to run using both Python 3.8 and whatever is the latest 3.x release (which at the time of writing is Python 3.12). To my surprise, [the job passes](https://github.com/nanopb/nanopb/actions/runs/10014462153/job/27684248087?pr=993).

@lubosz do you perhaps have a fresh install of a fast moving distro? Like Fedora?

lubosz:
@jaskij I do use Arch Linux, not a fresh install, but I can reproduce this in an virtualenv:
```
$ virtualenv fresh_env
created virtual environment CPython3.12.4.final.0-64 in 74ms
[...]

PR #993, opened 2024-07-19
run binary_packages on Python 3.8, and on latest 3.x

jaskij:
Due to #887 , it occurred to me it may be a good idea to test on the latest version of Python as well, among others to get early warnings of deprecations.

### DIFF
--- a/.github/workflows/binary_packages.yml
+++ b/.github/workflows/binary_packages.yml
@@ -12,10 +12,11 @@ on:
 jobs:
   build_linux:
     name: Build binary on Ubuntu 20.04
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         python-version: ['3.8', '3.x']
+        os: ['ubuntu-20.04', 'ubuntu-24.04']
 
     steps:
       - name: Check out code from GitHub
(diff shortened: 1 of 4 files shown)


jaskij:
@PetteriAimonen I managed to reproduce #887 in CI, and it rightfully causes the CI to fail. Do you want me to create a new PR with both this repro and the proposed solution from #994 ? Or how do you want me to proceed?

PetteriAimonen:
Yes, I think the CI test from this pull request and the fix from #994 would be good. I don't think the `check_pkg_resources.sh` is necessary if the CI build now catches this.

jaskij:
No, it won't be. That script was a workaround for CI not catching this.

jaskij:
@lubosz can you review the changes? It seems to be working now, but I'm not familiar enough with Python to be sure.

From what I read, this should work in the general case, but there may be some weird edge case with grpcio-tools being installed as an archive that I missed.

[CI run](https://github.com/nanopb/nanopb/actions/runs/10015300359/job/27686494415) showing that it does indeed catch #887

jaskij:
I talked the changes through with someone who has Python experience off GitHub, and he agreed this will work. The only breakage that could happen, which wouldn't with `pkg_resources` would be if `grpcio-tools` changes the way it's shipped, and it shomehow is an archive. Which is unlikely, and should be quickly caught by CI anyway.

PetteriAimonen:
Thanks a lot!

PR #994, opened 2024-07-19
Remove pkg resources

jaskij:
Disclaimer: I'm not familiar enough with Python to know that the changes I made are actually good. The tests past locally for me.

What this change does is abstract the two existing uses of `pkg_resources` into a single function, and inside it only uses that if Python is in version 3.8 or older.

The other option, which I didn't want to do, was to add a new dependency on [`importlib_resources`](https://pypi.org/project/importlib-resources/) for older Python versions.

3.9 is the cutoff, since it introduces the `importlib.resources` APIs this code used.

I also added a small script, `check_pkg_resources.sh`, which is a Bash one liner that checks there are exactly two instances of `pkg_resources` in all non-binary files under `generator/`.

### DIFF
--- a/.github/workflows/trigger_on_code_change.yml
+++ b/.github/workflows/trigger_on_code_change.yml
@@ -35,7 +35,8 @@ jobs:
 
       - name: Check for pkg_resources
         run: |
-          ! grep -RniI pkg_resources nanopb/generator
+          cd nanopb
+          ./check_pkg_resources.sh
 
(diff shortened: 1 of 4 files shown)


jaskij:
Note: since the behavior between `pkg_resources` and `importlib` has changed, this will most likely fail on some rare edge cases not captured in the CI.

jaskij:
Closing in favor of #993

PR #995, opened 2024-07-22
Fix Comment position with optional is used.

AkihiroUeda35:
If a field is optional and has a leading comment, leading comment is written before "has_field".

```proto
message Sample {
  // Comment
  optional int32value = 1;
}
```

```c
typedef struct _SampleMessage {
    /* Comment*/
    bool has_value;
    int32_t value;
} SampleMessage ;
```

In this condition, Intellisense comment of VS code is shown on has_value field.
With this change, comment position is on appropriate position as shown below.

```c
typedef struct _SampleMessage {
    bool has_value;
    /* Comment*/
    int32_t value;
} SampleMessage ;
```

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -774,7 +774,7 @@ class Field(ProtoElement):
         var_name = Globals.naming_style.var_name(self.name)
         type_name = Globals.naming_style.type_name(self.ctype) if isinstance(self.ctype, Names) else self.ctype
         leading_comment, trailing_comment = self.get_comments(leading_indent = True)
-
+        leading_comment_written = False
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
I'm not sure which is better. When reading code directly, it makes more sense to have the comment before `has_field`.

AkihiroUeda35:
Thank you for your comment.
Then, how do you think if generated code becomes as shown below?

```c
typedef struct _SampleMessage {
    /* Comment*/
    int32_t value;
    bool has_value;
} SampleMessage ;
```

PetteriAimonen:
That won't work, `has_value` has to precede the value for the descriptor offsets to remain in range.

PR #1003, opened 2024-08-28
Fix NANOPB_GENERATE_CPP_RELPATH being relative to the wrong directory

zachdeibert:
This fixes `NANOPB_GENERATE_CPP_RELPATH` being passed to `protoc` via `_nanopb_include_path` and `NANOPB_OPTIONS_DIRS` as a relative path.  `protoc` is ran in `CMAKE_CURRENT_BINARY_DIR` and `NANOPB_GENERATE_CPP_RELPATH` is relative to `CMAKE_CURRENT_SOURCE_DIR`, so the resulting path is incorrect.

In previous builds, the GitHub action [Build binary packages](https://github.com/nanopb/nanopb/actions/workflows/binary_packages.yml) prints `proto: warning: directory does not exist.` when building `examples/cmake_relpath`.  This should also fix that warning.


PetteriAimonen:
I must admit that I couldn't fully understand the problem scenario, but the changes seem reasonable.

Do you see any situation where this could break someones existing build?

zachdeibert:
The problem is that (using `examples/cmake_relpath` as an example) when `nanopb_generate_cpp` is called with `RELPATH proto`, that gets passed to `protoc` as `-Iproto --nanopb_opt=-Iproto`.  However, `protoc` is called from `examples/cmake_relpath/build`, not `examples/cmake_relpath`, so `protoc` interprets those paths as `examples/cmake_relpath/build/proto`, not `examples/cmake_relpath/proto`.  That's why it prints `proto: warning: directory does not exist.` (since `examples/cmake_relpath/build/proto` doesn't exist).

The only way it could break someone's build is if they were relying on this behavior and say `RELPATH proto` then put their `.proto` or `.options` files in the `build` folder (probably being generated by some other custom command).  In this case they would just have to update their call to `nanopb_generate_cpp` to say `RELPATH ${CMAKE_CURRENT_BINARY_DIR}/include`.  This seems reasonable though since every other CMake command uses paths relative to the current directory, not the `build` directory.

PetteriAimonen:
Thanks for the explanation!

PR #1008, opened 2024-09-18
wip generate message-specific decode and release routines

inolen:
Not actually proposing to merge this, but wanted to gather feedback on if this was useful and if so, how it should be implemented.

Decoding messages (specifically load_descriptor_values) was eating up the majority of my run time, so I wanted to take a stab at implementing decode / release using the existing field defines. I only implemented enough to get the messages I was working with decoded, which did shrink my message decoding overhead down to a negligible amount.


PetteriAimonen:
Yes, this is something I would eventually like to see included and the approach looks great.

I think this is going in a good direction, with a few things left to do:
* Tests
* Function for encoding
* Option so that this can be enabled per-message

I wonder whether it would make sense to combine all the optimized routines to a single function pointer or not. It would save a few words worth of function pointers (which otherwise have to be in every message) and also provide ABI compatibility if we later want to add more function types. On the other hand it is clearer if they are separate function pointers. The encode/decode would fit well with the current callback approach, but release does not.

inolen:
Awesome, I'll get the tests passing next.

RE: function pointers, I guess it's worth noting we don't _really_ need pb_decode / pb_release to internally be able to call these routines. I did it that way just for ease of testing, but it seems reasonable to just call the correct routine instead, especially since they don't need to take the pb_msgdesc_s struct anymore.

```
pb_decode(&stream, CMsgPlayerInfo_fields, &msg);
```
could just become
```
pb_decode_CMsgPlayerInfo(&stream, &msg);
```

inolen:
@PetteriAimonen wanted to ask one question as I'm closing in on _most_ of the tests passing. Right now I just started running all scripts using the "fast" path by running

```
scons CCFLAGS=-DPB_ENABLE_DECODE_FAST
```

I've had to hack and slash most of the SConscripts, as right now many of the binaries are only linked with pb_encode.o or pb_decode.o, as well many environments are created the use / don't use malloc.

This is problematic when code is generated directly in the *.pb.o, as it's not guaranteed to be linked with pb_decode.o, and only the pb_encode.o and pb_decode.o are generating malloc / non-malloc variants (we'd need variants also for each *.pb.o).

I think it's likely valuable that all tests past with the regular and "fast" paths, but as is that change feels very invasive to the test build scripts. Do you have any suggestions on the best way to go about this?

PetteriAimonen:
Indeed, that may be a bit difficult in current test structure.

I think for a start it could make sense to have a specialization of the `alltypes` test case for the fast path code, similar to how `tests/buffer_only` works. Perhaps also of the `fuzztest`.

inolen:
@PetteriAimonen extensions are working, but ran into one issue - is the maximum extension tag encoded somewhere i'm missing?

e.g. in alltypes.proto I see `extensions 200 to 255;` but then the FIELDLIST just contains the value 200.

PetteriAimonen:
No, only the start of the extension range is stored. Whether individual extension is recognized is then checked by the hooks.

inolen:
Just pushed the latest work, which is ready enough to get a solid review I think.

With this, all tests pass by just running `scons CCFLAGS="-DPB_ENABLE_DECODE_FAST"`. The interface from pb_decode.c is passed via a structure so that it works without modifying the tests. I think I want to add a PB_ENABLE_STATIC_DECODE_FAST / PB_ENABLE_DYNAMIC_DECODE_FAST to control that. Most users are going to want to use PB_ENABLE_STATIC_DECODE_FAST, but the dynamic option is nice for the tests right now.

I haven't actually worked on profiling and optimizing anything - it's just much faster for my use case because of the surface-level performance improvements. There's a couple of pieces (e.g. decoding callback fields) which I think are likely slower / blowing up code size by being inlined as they are. I'll start spending some cycles on profiling and hacking away at optimizing things now that everything is functional.

pb.h
```diff
@@ -344,6 +359,10 @@ struct pb_msgdesc_s {
 
     bool (*field_callback)(pb_istream_t *istream, pb_ostream_t *ostream, const pb_field_iter_t *field);
 
+    bool (*init)(pb_decode_if_t *, void *, unsigned int);
+    bool (*decode)(pb_decode_if_t *, pb_istream_t *, void *, unsigned int, uint32_t, pb_wire_type_t);
+    void (*release)(pb_decode_if_t *, void *, unsigned int, uint32_t);
+
```
PetteriAimonen, line 365: Could these be put into `PB_ENABLE_DECODE_FAST` ifdef? That would save 12 bytes of flash per message when the feature is not used.

pb_decode.c
```diff
@@ -164,25 +155,72 @@ pb_istream_t pb_istream_from_buffer(const pb_byte_t *buf, size_t msglen)
...
+}
+
+static void *pb_decode_realloc(void *ptr, size_t size)
+{
+    return pb_realloc(ptr, size);
+}
+
+#endif
+
+static void pb_decode_get_interface(pb_decode_if_t *pb) {
```
PetteriAimonen, line 176: I find this interface thing intriguing. On one hand it feels like it could have significant overhead, especially building it into RAM every time. On the other hand it could have some interesting optimization possibilities.

Could you elaborate why the functions are called through the interface pointers instead of directly?

pb_decode_fast.h
```diff
@@ -0,0 +1,1644 @@
...
+        {                                                                                                                                  \
+            if (!pb->decode_double_as_float(stream, (float *)data))                                                                        \
+                PB_RETURN_ERROR(stream, "failed to read wire value");                                                                      \
+        }                                                                                                                                  \
+        else if (pb->decode_fixed64)                                                                                                       \
+        {                                                                                                                                  \
+            if (!pb->decode_fixed64(stream, (double *)data))                                                                               \
+                PB_RETURN_ERROR(stream, "failed to read wire value");                                                                      \
```
PetteriAimonen, line 558: I think most of these `PB_RETURN_ERROR` could be just `return false`. The nanopb functions called should have set the error message already before returning false. In most cases, the error is coming from `pb_read()`.

`PB_RETURN_ERROR` does not overwrite existing error message, so there is no harm in this. But all these strings will use unnecessary memory and code size. It can also be confusing to follow, as you'd expect to get `"failed to read wire value"` if it fails here, but really you get the message from `pb_read()`.

pb_decode.c
```diff
@@ -294,9 +327,23 @@ bool checkreturn pb_decode_tag(pb_istream_t *stream, pb_wire_type_t *wire_type,
...
+        return false;
+    }
+
+    if (!pb_decode_varint32(stream, &temp))
+    {
+#ifndef PB_BUFFER_ONLY
+        /* band-aid for issue #1017. eof has been hit, but may have a lingering io error due to callbacks */
+        if (stream->callback != buf_read && stream->bytes_left == 0)
+        {
+            PB_SET_ERROR(stream, NULL);
```
PetteriAimonen, line 343: I think this is no-op because `PB_SET_ERROR` is:

     #define PB_SET_ERROR(stream, msg) (stream->errmsg = (stream)->errmsg ? (stream)->errmsg : (msg))

I think it is unnecessary, as the prior implementation didn't null out any error message either.

There is a subtle difference between this and the previous one: the previous one would only allow valid EOF at the first byte of the tag. But there is no guarantee about detecting corrupted messages anyway, so this is probably fine.

PetteriAimonen (commented):
I added some review comments.

I must say, looking at these macro constructions (both my own and now yours) always makes me both disgusted and impressed :)

pb_decode.c
inolen: It was really just a workaround for how the tests are compiled right now.

The tests recompile pb_encode.c / pb_decode.c with different options enabled / disabled, and often then only link a binary with one or the other + the *.pb.o files. This makes it hard not only to just get the executable to link (e.g. when only pb_encode.o is linked), but means the *.pb.c files would also need to be recompiled when toggling malloc, double to float conversion, etc.

Exporting the interface means the *.pb.o code just ends up consuming whatever is in pb_decode.o - and it doesn't matter if it doesn't exist. I'd like to had a DYNAMIC / STATIC switch for the fast decode, where the STATIC which directly calls pb_*().

PetteriAimonen: It seems like this is just an issue of the test framework currently. I don't like such detail being frozen into the API design. I'd rather just have more limited test coverage for this new feature for the time being.

pb_decode.c
inolen: Ah, right, I didn't actually test the functionality of canceling out the error. I just noticed that possibility while trying to track down issues that came up from my main loop not having a condition on bytes_left like the one in pb_decode.c

inolen:
Did a little bit of work on optimizing some of the generated code and I think a fairly common pattern is emerging.

There are a number of pb_dec_* routines which right now take in a stream and pb_field_iter_t pointer. I think most of these can be split such that we end up with something like -

```
static bool checkreturn pb_dec_string_raw(pb_istream_t *stream, void *dest, size_t capacity)
{
   ...
}

static bool checkreturn pb_dec_string_field(pb_istream_t *stream, const pb_field_iter_t *field)
{
    if (PB_ATYPE(field->type) == PB_ATYPE_POINTER)
        return pb_dec_string_raw(stream, field->pData, 0);
    else
        return pb_dec_string_raw(stream, field->pData, field->data_size);
}
```

where pb_dec_*_raw can be used by the fast path, which is just stripping all of the logic that exists inside pb_dec_*_field accessing the generated field data variant due to the specialization.

PetteriAimonen:
I'm a bit worried that in trying to heavily optimize for speed, there will be significant code size penalties even when not using this feature. Nanopb primarily tries to be a small size library, not necessarily high performance library.

inolen:
> I'm a bit worried that in trying to heavily optimize for speed, there will be significant code size penalties even when not using this feature. Nanopb primarily tries to be a small size library, not necessarily high performance library.

Got it. I'll keep an eye to not bloat that + make sure everything possible is getting ripped out when disabled.

inolen:
Finished the work to merge the lvalue decoding logic between pb_decode.c and pb_decode_fast.h.

The size of the pb_decode.o compiled by the unit tests is a ~4 kb larger due to some of the added functions, but the final decode_alltypes executable (as a reference) is only 1kb larger.

I could probably trim that down further, but I don't know how important it is (vs the code changes). The biggest code blow up / change here was splitting pb_dec_varint into 3 seperate routines for varint, svarint and uvarint. I changed the logic for how overflows are detected to trim down the code size there in order to mostly offset the added routines. Rather than depending on casts and temporary variables, it's now shifting the decoded value to manually check for the overflow in the high bits.

I also at the end added a flag for `PB_DYNAMIC_DECODE_FAST` and made the default route statically link with the various routines from pb_decode.c. Having the global interface struct isn't ideal, but it removes it from being baked into the API. From debugging the code during bring up, I'd argue it's worth leaving in even if it's a bit ugly so that all tests can easily be ran through the fast path.

All tests should run with `scons CCFLAGS="-DPB_DECODE_FAST -DPB_DYNAMIC_DECODE_FAST"` as well as just `scons` now.

PetteriAimonen:
Thanks, this is looking quite ready.

I'd prefer if you can rewrite the commit history a bit:
* Changes to existing code before adding the new feature (if possible, if there are a lot of conflicts no need to)
* Each commit should have a reasonable commit message by itself.

inolen:
Will do. Do you have any feelings regarding the file names / preprocessor defines?

I went with PB_DECODE_FAST at first, and then I tacked on PB_DYNAMIC_DECODE_FAST but I didn't put too much thought into it. Perhaps it'd make more sense to have one explicitly say static, and one explicitly say dynamic. However, the dynamic route may just get axed down the line once the tests get updated and then having one explicitly say static is odd.

PetteriAimonen:
Thanks, I will be merging these one-by-one as I have time to review.

Don't mind the conflict warnings.

inolen:
@PetteriAimonen thanks a lot! Please let me know if there's any follow up tasks you'd like to see done, etc.

inolen:
Started pushing a few wip experiments to that branch that aren't required, but I thought I'd point out that https://github.com/nanopb/nanopb/pull/1008/commits/515914eff6f16bf7a12fac4d1e8d033736f749e1looks like a legit bug in both code paths.

If pb_close_string_substream fails after successfully decoding, pb_release isn't called.

inolen:
Ripped out my WIP changes to a new branch, and pushed both that fix for pb_release and another change to directly call a submessage's decode handler.

inolen:
@PetteriAimonen ripped out the memcpy and replaced with an if conditional. One thing I didn't understand when I was reverting that code - why did the previous code cast to the *int_least*_t types when writing out the value? Couldn't that ultimately result in writing out more than intended on platforms where sizeof(*int_least*_t) != sizeof(*int*_t).

PetteriAimonen:
@inolen The field size is tested against `int_least*_t` so the size will be correct. The reason for these is platforms that do not have 8- or 16-bit types.

inolen:
@PetteriAimonen ah, correct, data_size is compared against that. So forgive my ignorance, do these platforms not have *int16_t and *int8_t defined then?

PetteriAimonen:
Yeah, there are platforms where the smallest addressable memory unit is 16 or 32 bits. This is somewhat common on older DSP processors, such as C54x. Nanopb has had support for them for a while, but the test coverage is quite sparse as I don't know of a suitable simulator.

PR #1010, opened 2024-09-19
update descriptor.proto to latest

Sayter99:
The descriptor.proto lives in this repo https://github.com/nanopb/nanopb/blob/master/generator/proto/google/protobuf/descriptor.proto is 6-year old, I wonder if it is possible to update it so that we can upgrade protobuf version for bazel.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -4,10 +4,10 @@ module(
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.7.1")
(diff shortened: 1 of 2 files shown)


Sayter99:
some CI fails with this message `Error: This request has been automatically failed because it uses a deprecated version of "actions/upload-artifact: v2"`; is it known issue?

PetteriAimonen:
Yeah, it started doing that this week. I need to see about updating the CI.

PR #1016, opened 2024-09-30
Various stack size reduction tweaks

petejohanson:
As part of my work integrating nanopb into ZMK (by way of the Zephyr integration), like many others I've been hit with some high stack usage. The attached changes helped drop the usage by about 300bytes with my testing. The current ZMK PR integrating this change is up at https://github.com/zmkfirmware/zmk/pull/2505

* Add option for skipping size checks when encoding submessages
* Tweak inline/noinline for functions to reduce stack size.

Not sure if this is something you'd be interested in, but I wanted to at least propose upstream here. Thanks for the project!

### DIFF
--- a/pb.h
+++ b/pb.h
@@ -28,6 +28,9 @@
 /* Disable support for error messages in order to save some code space. */
 /* #define PB_NO_ERRMSG 1 */
 
+/* Disable checks to ensure sub-message encoded size is consistent when re-run. */
+/* #define PB_NO_ENCODE_SIZE_CHECK 1 */
+
 /* Disable support for custom streams (support only memory buffers). */
 /* #define PB_BUFFER_ONLY 1 */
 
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
Very nice! I find it surprising that these small changes have such a large effect.

PetteriAimonen:
/home/runner/work/nanopb/nanopb/nanopb/pb.h:132:1: error: C++ style comments are not allowed in ISO C90
      132 | // Define for explicitly not inlining a given function

This needs to be changed, otherwise looks ready to merge.

I think I'll try to rework the `PB_NO_ENCODE_SIZE_CHECK` to be as efficient while keeping the check, but for now this is ok.

petejohanson:
> ```
> /home/runner/work/nanopb/nanopb/nanopb/pb.h:132:1: error: C++ style comments are not allowed in ISO C90
>   132 | // Define for explicitly not inlining a given function
> ```
> 
> This needs to be changed, otherwise looks ready to merge.
> 
> I think I'll try to rework the `PB_NO_ENCODE_SIZE_CHECK` to be as efficient while keeping the check, but for now this is ok.

Fixed. Thanks!

PetteriAimonen:
Merged, thanks!

PR #1026, opened 2024-10-23
Support custom naming styles

joshstrohminger:
This is an attempt to support custom naming by loading a python module via a module path and class name. I'm happy to hear feedback and suggestions for alternate approaches. I added a test named _namingstyle_custom_, which was copied and modified from _namingstyle_ to use fewer delimiters and more camel and pascal casing.

## Passing in the styler (fourth commit)

I tried to make this more flexible by also passing a class name but it could be simplified by requiring a specific class name if that's preferred. I first tried to pass the arguments in as a tuple (`nargs=2` in `optparser.add_option`) but that didn't seem to play nice when trying to run the tests or running it as a plugin; it kept getting split. I could wrap it in single quotes but then I'd have to strip those which felt wrong, so I switched to passing a single delimited string. A comma seemed appropriate as the delimiter but that didn't play nice in all cases either, so I switched to a dot since passing the option in the form _MODULE.CLASS_ felt better.

The new _--custom-style_ option is manually checked for being mutually exclusive with the _-C/--c-style_ option. I see that _arpparse_ supports this but _optparse_ doesn't seem to. I could create a separate PR to migrate to using _argparse_ if there is any interest.

The value is split via a callback so the parsing/validation can stay within _optparser_. It's then interpreted as being a file path, optionally with a `".py"` extension. The module name is simply the basename without the extension.

The module is loaded using the approach from [an example in the python docs](https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly) and put into a local class so that _NamingStyle_ is used as the last inherited class (or _NamingStyleC_ if the _-C/--c-style_ flag is also used), allowing the custom styler to only implement the methods they want to modify.

## Modification to support more styling (all commits)

When using a custom styler I noticed a few places where I had issues with styling. Note that all tests still pass after these changes and I only saw one difference in the generated code (see below).

The reverse name mapping was affected when there was a package name that was stripped via mangling. It would generate mappings to things that were part way between being properly styled and canonical. To remedy this, I checked if the styled name was different and if the canonical name was different (to match a check that was being used when yielding the mapping definitions. I think this only affects custom stylings but it ends up with a mapping to a mapping. Here's an example from _custom_naming_style_mangle.pb.h_ that is generated from the new test:

```c
/* Mapping from canonical names (mangle_names or overridden package name) */
#define package_MY_ENUM2 MY_ENUM2
#define MY_ENUM2 MyEnum2
```

If I didn't to the double mapping, generation would fail elsewhere because it would fail to lookup enums from the dependencies.

43140b349631db814588aa8d12252cdc444aaed6 now simply uses the last part of the long names for each enum instead of trying to strip the type name prefix from it, which was assuming there was a single `_` as the delimiter. This avoids that assumption.

af7ec95dd18021b8dee09e41ba4e28716c53845a removes a double-styling that was occurring for static bytes types. It was getting styled by calling `var_name`, which didn't feel quite right to me since it was a `typedef`, but the underlying `ctype` already gets styled by calling `bytes_type(...)` in the `Field.__init__` method: https://github.com/nanopb/nanopb/blob/af7ec95dd18021b8dee09e41ba4e28716c53845a/generator/nanopb_generator.py#L746
This happened to work out because the styling was effectively the same, but would be a problem for a custom styler.

daa170bd9694bcf703afccb998bb3aa8b45d4717 styles the extension fields as variables. This only affects the generated output when using the _-C/--c-style_ flag. You can see a diff below of _naming_style.pb.c_ that gets generated from the _namingstyle_ test:

```diff
 #define TEST_EXTENSION_TEST_EXTENSION_EXTMSG_CALLBACK NULL
 #define TEST_EXTENSION_TEST_EXTENSION_EXTMSG_DEFAULT NULL
 #define test_extension_test_extension_extmsg_t_test_extension_MSGTYPE test_extension_t
-pb_byte_t TestExtension_testExtension_extmsg_default[] = {0x00};
+pb_byte_t test_extension_test_extension_extmsg_default[] = {0x00};
 PB_BIND(TEST_EXTENSION_TEST_EXTENSION_EXTMSG, test_extension_test_extension_extmsg_t, 2)
 const pb_extension_type_t test_extension_test_extension = {
     NULL,
```

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -166,6 +166,7 @@ There are also a few command line options that cannot be applied using the
 other mechanisms, as they affect the whole generation:
 
 * `--c-style`: Modify symbol names to better match C naming conventions.
+* `--custom-style`: Modify symbol names by providing your own styler implementation.
 * `--no-timestamp`: Do not add timestamp to generated files.
 * `--strip-path`: Remove relative path from generated `#include` directives.
 * `--cpp-descriptors`: Generate extra convenience definitions for use from C++
@@ -224,6 +225,12 @@ For a full list of generator command line options, use `nanopb_generator.py --he
                             Include insertion point comments in output for use by
                             custom protoc plugins
     -C, --c-style         Use C naming convention.
+    --custom-style=MODULE.CLASS
+                          Use a custom naming convention from a module/class
+                          that defines the methods from the NamingStyle class to
+                          be overridden. When paired with the -C/--c-style
+                          option, the NamingStyleC class is the fallback,
+                          otherwise it's the NamingStyle class.
 
     Compile file.pb from file.proto by: 'protoc -ofile.pb file.proto'. Output will
     be written to file.pb.h and file.pb.c.
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1856,6 +1856,11 @@ class MangleNames:
             self.name_mapping[str(names)] = new_name
             self.reverse_name_mapping[str(new_name)] = self.canonical_base + names
 
+            styled_name = Globals.naming_style.type_name(new_name)
+            if str(new_name) != str(self.canonical_base + names) and styled_name != str(new_name):
+                # If a custom styling doesn't match the canonical mangled name, which also doesn't match the mangled name, add a reverse mapping between them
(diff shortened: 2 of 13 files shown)


PetteriAimonen:
Thanks, excellent work!

PR #1034, opened 2024-11-22
Update grpcio-tools version in requirements.txt

tpudlik:
This fixes errors of the form `ImportError: cannot import name 'runtime_version' from 'google.protobuf'` when building nanopb protos with Bazel.

Since https://github.com/nanopb/nanopb/pull/1010, nanopb depends on protobuf 5.28.2. However, nanopb also depends on the protobuf pip package, [via
grpcio-tools](https://github.com/nanopb/nanopb/blob/48ceadfdf1e6446f073e2cae83f18fe78861e0e9/extra/requirements_lock.txt#L119-L133). Unfortunately, grpcio-tools 1.59.3 pull in python-protobuf 4.22.0, which is incompatible with the 5.82.2 protoc
(https://stackoverflow.com/a/78747599/24291280). This leads to the `ImportError` above if `google.protobuf` ends up being imported from the pip-managed python-protobuf package, but `nanopb_pb2.py` was generated using protobuf 5.82.2.

This PR updates grpcio-tools to the newest version, which updates python-protobuf to 5.82.3.

This should have been done as part of https://github.com/nanopb/nanopb/pull/1010 (and so made it to the 0.4.9 release).


tpudlik:
Example failure: https://pwbug.dev/354274498#comment12 (with some discussion at https://pwbug.dev/354274498#comment23).

PetteriAimonen:
Thanks! Do you have ideas how we could improve the CI testing to catch problems like this?

tpudlik:
I thought about how to create a regression test for this, but as I did so, I realized that the whole issue only arises because of an unnecessary dependency edge. So I sent another PR to cut it!

PR #1045, opened 2024-12-11
Handle case when PB_NO_ENCODE_SIZE_CHECK is not defined

twam:
* Handle PB_NO_ENCODE_SIZE_CHECK like other defines, so that it also builds correctly when -Wundef is set


twam:
@petejohanson Maybe you can have a look here, as the original changed was introduced by you.🙏

PetteriAimonen:
Looks good, I also added `-Wundef` to tests. Apparently that is not included in `-Wall -Wextra`, which was pretty surprising.

twam:
Was also surprised at first, that `-Wundef` is not part of `-Wall -Wextra` when analysing the issue in our setups 😁

Issue #1041, opened 2024-12-01
Automate release workflow, switch to GitHub releases


PetteriAimonen:
Nanopb has been using custom hosted download server as back when it was set up, GitHub didn't have releases functionality yet.

Nowadays it would simplify release management to have them as GitHub releases instead.

tpudlik:
As part of the release workflow, could [the version in MODULE.bazel](https://github.com/nanopb/nanopb/blob/master/MODULE.bazel#L3) be incremented? I noticed this was not done for the 0.4.9.1 release. I patched this line when adding the release to the Bazel Central Registry, but it would be nice for it to be correct in the source.

PetteriAimonen:
Yeah, that should be added to `set_version.sh`: https://github.com/nanopb/nanopb/blob/master/tools/set_version.sh

weichslgartner:
Could the version information also be added to `zephyr/module.yml` (see https://docs.zephyrproject.org/latest/develop/modules.html#vulnerability-monitoring). This would enable to generate SBOMs in zephyr projects using nanopb.

For example adding CPE and PURL for version 0.4.0 would look like:
```yml
security:
  external-references:
    - cpe:2.3:a:nanopb_project:nanopb:0.4.0:*:*:*:*:*:*:*
    - pkg:github/nanopb/nanopb@V0.4.0
```

PetteriAimonen:
@weichslgartner Sure, can you make a pull request for set_version.sh modification?

weichslgartner:
@PetteriAimonen Sure, will create an PR.

pdgendt:
Any ETA on the 1.0.0 release?

PetteriAimonen:
Probably many months, I haven't done much towards it.

PR #1048, opened 2024-12-17
Set version in MODULE.bazel

tpudlik:
Tested by running the script locally.

Addresses https://github.com/nanopb/nanopb/issues/1041.


PetteriAimonen:
Thanks! The more the merrier, I'm sure one day we'll have perl in the script, too :)

PR #1062, opened 2025-02-25
PB_INLINE_CONSTEXPR for fields() has_msg_id() and msg_id()

Nizerlak:
`inline`s alone for the static functions of `MessageDescriptor`s are sometimes not enough for some use cases. These changes make them more useful.


PetteriAimonen:
Looks ok to me.

@denravonska Do you want to take a look?

denravonska:
Not tested but it makes sense :)

PetteriAimonen:
Thanks both :)

Nizerlak:
Thank you :)

Issue #1061, opened 2025-02-24
Decoding submsg_callback with oneof FT_CALLBACK fails silently


HarshaLaxman:

PetteriAimonen:
Yeah, this is a common problem people run into, for which there are various workarounds but no universal solution.

From [docs](https://jpa.kapsi.fi/nanopb/docs/concepts.html#oneof):
> When a field inside oneof contains pb_callback_t fields, the callback values cannot be set before decoding. This is because the different fields share the same storage space in C union. Instead either function name bound callbacks or a separate message level callback can be used. See [tests/oneof_callback](https://github.com/nanopb/nanopb/tree/master/tests/oneof_callback) for an example on this.

You can also consider using `no_union` option. Or of course, if you know the maximum string length, just set it to the generator `max_size` option and avoid callbacks altogether.

PetteriAimonen:
Ah, sorry, I didn't read your code carefully enough.

Indeed this feels like an omission, as currently the submessage callback is only invoked when the submessage itself is static, and callbacks are only used inside of it. I guess it would be possible to copy the logic from `pb_enc_submessage` / `pb_dec_submessage` to `pb_default_field_callback`.

Can you elaborate on the issues you are having with name-bound callbacks? I hope that eventually I would be able to come up with a better solution for this long standing issue, so I'm trying to understand the practical problems.

HarshaLaxman:
Name bound callbacks feel shaky

For example, iiuc in the [tests/oneof_callback](https://github.com/nanopb/nanopb/tree/master/tests/oneof_callback), adding another message would cause the `SubMsg3_callback` to be ran for `AnotherMessage` also, with no additional code.
```
message AnotherMessage
{
    option (nanopb_msgopt).submsg_callback = true;

    oneof values
    {
        int32 intvalue = 1;
        SubMsg3 submsg3 = 2;
    }
}
```

Even if there is a way to add code to check which top level message the field belongs to in `SubMsg3_callback`, it's a bit of hidden logic.

It's preferable to explicitly set the callbacks from the top level message on down.

HarshaLaxman:
Put up a PR with a fix and regression test. All tests pass, but it's likely I missed some edge cases

https://github.com/nanopb/nanopb/pull/1063/files

PR #1063, opened 2025-02-25
Fixes callback field within submsg - issue #1061

HarshaLaxman:


pb_decode.c
```diff
@@ -782,6 +782,22 @@ static bool checkreturn decode_callback_field(pb_istream_t *stream, pb_wire_type
         
         if (!pb_make_string_substream(stream, &substream))
             return false;
+
+        /* If the callback field is inside a submsg, first call the submsg_callback which
```
PetteriAimonen, line 786: All submessages don't have a submessage callback before them, so it should check for `SUBMSG_W_CB` and generator should be changed to put that.

HarshaLaxman: Updated

HarshaLaxman: @PetteriAimonen ?

PetteriAimonen: I haven't had time to review this.

PetteriAimonen:
Thanks, looks ok.

There was a regression in a special corner case of callback fields outside of oneofs when a message level option is specified.
Pretty niche, but I fixed that anyway and will merge once CI checks complete.

PR #1089, opened 2025-05-14
Add virtual env for generator in example/simple

cgringmuth:
I added some convenience for easier experimenting with simple example.

Maybe it helps other devs as well.

PS: Sorry for earlier PR, it contained wrong author.

### DIFF
--- a/examples/simple/README.txt
+++ b/examples/simple/README.txt
@@ -14,11 +14,26 @@ The code uses four different API functions:
 Example usage
 -------------
 
-On Linux, simply type "make" to build the example. After that, you can
+
+Linux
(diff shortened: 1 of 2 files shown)


PetteriAimonen:
As this currently is, I'm not going to merge it.

Using virtualenv is a reasonable way to install the dependencies, and towards that end a `requirements.txt` already exists in the main folder (and weirdly another one in extra folder). Adding yet another doesn't seem like a good idea.

I agree that there could be more documentation regarding installation on various platforms and in various ways, either in README or in the docs. I don't think installation documentation belongs in the README of an example.

PR #1090, opened 2025-05-15
Compute size of all statically known messages

schultetwin:
Fields that have a user defined different type (such as a FT_POINTER) may still have a known encoded size.

This pull requests computes the encoded sizes of those messages to ensure that, if nanopb has enough information to know the encoded size, it is reported.


PetteriAimonen:
Looks reasonable, but some of the test cases probably need updating.

schultetwin:
Thanks @PetteriAimonen. I forgot to `rm -rf build` after making some edits to re-run tests. I think it's fixed now!

PR #1107, opened 2025-07-22
add support for custom `noinline` definition

bonifaido:
I have a gcc compiled environment, where `inline` is defined as a macro already when compile hits nanopb, and thus compilation fails. The aforementioned macro definition is like the one here: https://elixir.bootlin.com/linux/v6.15.7/source/include/linux/compiler_attributes.h#L231

This PR checks if `noinline` is "redefined" as a macro and skips creating a new definition for it.


PetteriAimonen:
Thanks!

PR #1113, opened 2025-09-03
Add possibility to specify default encoder and decoder functions for string fields

T3RR7:
This PR introduces a feature that allows specifying default encoder and decoder functions for strings.

If either `default_string_decoder` or `default_string_encoder` is specified, additional initialization macros will be generated with the corresponding names: `*_zero_for_decode` or `*_zero_for_encode`.

An additional entity, `FieldEncodingOptions`, has been added to support further improvements in this area. For example, it will allow users to define encoder and decoder functions for repeated fields or arrays.

UPD:
PR also adds possibility to specify `PB_ENCODERS_HEADER` and `PB_DEFAULT_STRING_ENCODER` and `PB_DEFAULT_STRING_DECODER`

### DIFF
--- a/pb_common.c
+++ b/pb_common.c
@@ -323,11 +323,11 @@ bool pb_default_field_callback(pb_istream_t *istream, pb_ostream_t *ostream, con
 
             if (PB_LTYPE(field->type) == PB_LTYPE_STRING) {
             	if (decode == NULL) {
-            		decode = &PB_DEFAULT_STRING_DECODER;
+            		decode = PB_DEFAULT_STRING_DECODER;
             	}
 
             	if (encode == NULL) {
-            		encode = &PB_DEFAULT_STRING_ENCODER;
+            		encode = PB_DEFAULT_STRING_ENCODER;
             	}
             }
 


PetteriAimonen:
Thanks for your contribution. Unfortunately because of maintenance burden and lack of time, I'm not sure I can take it into the main nanopb codebase.

Could you describe the situation where this feature is useful?

T3RR7:
> Could you describe the situation where this feature is useful?

Of course.
This is useful when we share a memory buffer between two processes, or between kernel and user space.

[1st feature](https://github.com/nanopb/nanopb/pull/1113/files?utm_source=chatgpt.com#diff-e210870c197f23a0840de87fbe37e0803aaa4c48ac5108fb14e2bcee8855199bR57)
 with [default string encoder/decoder initialization](https://github.com/nanopb/nanopb/pull/1113/files?utm_source=chatgpt.com#diff-a01364727bf4728eb34beeece5624f9f5ff3b3e433ecec1d1caa40315df05269R2159)
 is needed for simple messages without nested structs or unions. It is very useful when, on the receiver side, we want to decode all strings with a uniform decoder. For example, a message like:

```protobuf
message Foo {
     string str1 = 1;
     string str2 = 2;
     string str3 = 3;
     string str4 = 4;
     string str5 = 5;
     // ...
     string str100 = 100;
}
```

would require a huge amount of boilerplate code like:

```c
Foo foo = Foo_init_zero;
foo.str1.funcs.decode = some_decode_func;
foo.str2.funcs.decode = some_decode_func;
/* for all 100 strings */
```

while it can be achieved with the generated code I implemented:

```c
Foo foo = Foo_init_zero_for_decode;
```

[The 2nd change](https://github.com/nanopb/nanopb/pull/1113/files?utm_source=chatgpt.com#diff-e210870c197f23a0840de87fbe37e0803aaa4c48ac5108fb14e2bcee8855199bR26)
 solves the problem when we want to work with more complex messages containing unions. It does the same thing as the first, but works a [bit more at runtime](https://github.com/nanopb/nanopb/pull/1113/files?utm_source=chatgpt.com#diff-873ad415246aa8ebaa8889fced3223f6c93588a7482408cff9861475094bea46R324)
.

For example, consider a proto like:

```protobuf
message Foo {
    string foo_str = 1;
}

message Bar {
    int32 bar_int = 1;
    string bar_str = 2;
}

message FooOrBar {
    oneof data {
        Foo foo = 1;
        Bar bar = 2;
    }
}
```

The init macros for FooOrBar cannot decide which field to initialize and will just place one of the initializers from oneof. But if we want to define a unified rule for string decoding without huge amounts of ugly code (and of course, without modifying the proto), we can simply specify default decoders that will be applied at runtime if no decoder is specified manually.

This is achieved by [defining PB_ENCODERS_HEADER](https://github.com/nanopb/nanopb/pull/1113/files?utm_source=chatgpt.com#diff-873ad415246aa8ebaa8889fced3223f6c93588a7482408cff9861475094bea46R8)
, which is an elegant trick IMO, similar to how PB_SYSTEM_HEADER works.

Coming back to my initial example: when we have a shared buffer, we can decode a string without any allocations, just by pointing to the string in the buffer. The current nanopb implementation doesn’t allow this with just a few lines of code, and the suggested changes fix that.

So now one the possible usage could be:

```c
/* encoders.h */

typedef struct ipc_string_field {
        const uint8_t *data;
	uint64_t size;
} ipc_string_field_t;

bool ipc_decode_string(pb_istream_t *stream, const pb_field_t *field, void **arg);

bool ipc_encode_string(pb_ostream_t *stream, const pb_field_iter_t *field, void *const *arg);

#define PB_DEFAULT_STRING_DECODER ipc_decode_string
#define PB_DEFAULT_STRING_ENCODER ipc_encode_string 
```

```c
/* encoders.c */

bool ipc_decode_string(pb_istream_t *stream, const pb_field_t *field, void **arg)
{
	*arg = stream->state;
	return true;
}

bool ipc_encode_string(pb_ostream_t *stream, const pb_field_iter_t *field, void *const *arg)
{
	const ipc_string_field_t *const str = *arg;

	if (str == NULL) {
		return true;
	}

	if (!pb_encode_tag_for_field(stream, field)) {
		return false;
	}

	/* with zero included to allow zero copy reference to buffer memory while decoding */
	return pb_encode_string(stream, str->data, str->size + 1);
}
```

```makefile
my_target_cflags += -DPB_ENCODERS_HEADER="<encoders.h>"
```

PetteriAimonen:
This could probably be achieved without modifications to the core code.
Something like this could work:

1. Message option `(nanopb).callback_function = "custom_string_callback"`
2. A custom callback based on `pb_default_field_callback` that checks `PB_LTYPE` of `field->type` and decides which function to call. For other than string fields it can delegate to the default callback.

PR #1115, opened 2025-09-15
Add container size define for array and string fields in nanopb generator

kay0u:
## Summary

This PR modifies the nanopb generator to use define for container sizes instead of hardcoded values in array declarations.

## Changes

  - Modified array declarations to use `%s_%s_container_size` macro pattern instead of
  hardcoded `[%d]` values
  - Added `container_size()` method to generate #define statements for container sizes       
  - Applied changes to:
    - Repeated fields with max_count
    - String fields with max_size
    - Bytes fields with max_size

  The generator now produces:
  - Container size definitions: #define MESSAGE_FIELD_CONTAINER_SIZE `<value>`
  - Array declarations using these macros: `field_type
  field_name[MESSAGE_FIELD_CONTAINER_SIZE]`

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -634,7 +634,7 @@ class Field(ProtoElement):
             if self.max_count is None:
                 self.can_be_static = False
             else:
-                self.array_decl = '[%d]' % self.max_count
+                self.array_decl = '[%s]' % ('%s_%s_container_size' % (self.struct_name, self.name))
                 if field_options.fixed_count:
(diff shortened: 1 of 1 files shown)


kay0u:
Indeed, I was unaware of the existence of this macro. However, with my solution, it is done at compile time.

Please feel free to close this PR if you are not interested.

PetteriAimonen:
The `sizeof()` runs at compile time too.

I agree that having the defines can sometimes make sense, but at this time I'm not going to merge this.

PR #1120, opened 2025-09-30
Use styled naming when generating c++ lookup code

Brambovich:
When i was using the C++ lookup generation, I saw that the naming style is hardcoded to be only the "default" style. 
While when a user applies the c-naming style and the C++ lookup by using the command line arguments: `--cpp-descriptors -C`, the naming is not correct in the generated C++ code, giving a compiler error. 

I'm not sure if I overlooked something, this is my first PR.


PetteriAimonen:
Thanks!

PR #1123, opened 2025-10-28
Meson support

Teselka:
This adds simple meson support to the nanopb with generator exposing support.

Also an example project how end-user would use it:
[nanopb_meson.zip](https://github.com/user-attachments/files/23189668/nanopb_meson.zip)

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -1,0 +1,52 @@
+name: Test Meson-based installation and compilation
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  push:
+    paths:
+      - '**meson**'
+  pull_request:
+    paths:
+      - '**meson**'
+
+jobs:
+  build_meson_linux:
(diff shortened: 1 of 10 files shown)


PetteriAimonen:
Can you add the example into the repository, and as an automatic test?

For example, add it as `examples/meson` and then add a new workflow under `.github/workflows` that tests it. The existing [cmake.yml](https://github.com/nanopb/nanopb/blob/master/.github/workflows/cmake.yml) and [platformio_tests.yml](https://github.com/nanopb/nanopb/blob/master/.github/workflows/platformio_tests.yml) are good examples to follow. Add the new workflow also into `trigger_on_schedule.yml` so that it is run every week.

If you enable the workflows in your own repository, you can test that the test works.

The version number in the meson file should be added to `tools/set_version.sh`, so that it is updated automatically to match the nanopb version.

PetteriAimonen:
Thanks, this looks good! I think it is reasonable to use `add_global_arguments()`.

Issue #1131, opened 2026-01-18
Type use before definition when generating with  "package" option


JeanTie:
At this point i am unsure if this is intended and i use it incorrectly or if this is a bug.

I stumbled upon this when trying to use a shared .proto file i must use defined by someone else.
To assure there will be no naming collision the package option is used. But this leads to the case where a type is used, before it is defined.

I am using version 0.4.9.1

A simple example:

**Steps to reproduce the issue**
`package_test.options`:
```
# Define package option to generate types with this package name as prefix.
* package:"ABC"
```

`package_test.proto`:
```
// A very simple definition (based on the simple example) to show the issue
syntax = "proto2";

enum ErrorCode {
    UNKNOWN = 0;
    SPECIFIED = 1;
}

message Error {
    required ErrorCode code = 1;
    optional string description = 2;
}
```

`python generator/nanopb_generator.py package_test.proto`

**What happens?**
We generate types like:

```c
/* Enum definitions */
typedef enum _ABC_ErrorCode {
    ABC_ErrorCode_UNKNOWN = 0,
    ABC_ErrorCode_SPECIFIC = 1
} ABC_ErrorCode;

/* Struct definitions */
typedef struct _ABC_Error {
    ErrorCode code;
    pb_callback_t description;
} ABC_Error;
```

**What should happen?**
When we expect:

```c
/* Enum definitions */
typedef enum _ABC_ErrorCode {
    ABC_ErrorCode_UNKNOWN = 0,
    ABC_ErrorCode_SPECIFIC = 1
} ABC_ErrorCode;

/* Struct definitions */
typedef struct _ABC_Error {
    ABC_ErrorCode code;
    pb_callback_t description;
} ABC_Error;
```

Difference beeing:

```diff
typedef struct _ABC_Error {
-   ErrorCode code;
+   ABC_ErrorCode code;
    pb_callback_t description;
} ABC_Error;
```

Even though the type is defined later in the file via macro:
```c
#define ErrorCode ABC_ErrorCode
#define Error ABC_Error
```
 We still have the type use before definition at the top.

---

If this does turn out to be a bug, I would be willing to take on the challenge to try and fix it. While I'm not very familiar with this project or the underlying technology, I'm eager to learn and contribute. I appreciate any guidance you can provide as I navigate this.

PetteriAimonen:
Indeed it is a bug, and thanks for the good example case.

[...]

PR #1132, opened 2026-01-18
Fix for issue #1131

JeanTie:
This fixes the code generation:

```diff
typedef struct _ABC_Error {
-   ErrorCode code;
+   ABC_ErrorCode code;
    pb_callback_t description;
} ABC_Error;
```

Thanks to your feedback @PetteriAimonen i checked `mangle_field_typename` to see how it handles the name mangling.\
Additionally your findings of a correctly generated source when defining package in the .proto file was interesting and proofed usefull.

What i found is:
* When definining package in the .proto file:
  * `canonical_base` is set here
* When not defining package in the .proto file:
  * `canonical_base` is empty
* When defining package in the .options file
  * `replacement_prefix` is still set

This patch leverages these findings for a simple fix by just checking if `replacement_prefix` is None. In the case it is not (e.g. package definition in .options file) we do not return typename, but let the existing logic handle the replacement.

Since there is still a ton of code that i haven´t even read yet, i am unsure if this is the best way to fix this.

This fix addresses the issue, but i am open to feedback on wether this is the best (or at least acceptable) approach.\
I´d appreciate any feedback and insight on this solution or suggestions for improvement.

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -1886,7 +1886,7 @@ class MangleNames:
             return "." + typename.split(".")[-1]
 
         canonical_mangled_typename = str(Names(typename.strip(".").split(".")))
-        if not canonical_mangled_typename.startswith(str(self.canonical_base) + "_") and (self.replacement_prefix is None):
+        if not canonical_mangled_typename.startswith(str(self.canonical_base) + "_") and self.canonical_base != Names():
             return typename
 
         if self.strip_prefix is not None and typename.startswith(self.strip_prefix):


JeanTie:
> Looks like it breaks some of the other tests. Otherwise I'm fine with the approach, but it appears it doesn't work for all cases.
> 
> Some thoughts:
> 
>     * The comment header in `package_name_prefix.proto` is not needed and is just confusing, the file can just have `* package: "ABC"` as the contents.
> 
>     * I'm not sure, but I wonder in the `canonical_mangled_typename.startswith(str(self.canonical_base) + "_")` comparison, instead of checking `replacement_prefix` it should just check for `self.canonical_base` to be empty?

Good  points. Sorry to require that much support, i am propably too deep in very unfamiliar technology here.
About the tests - how am i supposed to run them?
I tried:
* make the generator package via the tools/scripts
* Run `scons` in the tests directory
But even when i run it without my changes on the main branch it fails early with:
```
build\message_sizes\dummy.c(9): error C2065: "WithPointerSubmsg_size": undeclared identifier
build\message_sizes\dummy.c(9): error C2057: expected constant expression
scons: *** [build\message_sizes\dummy.obj] Error 2
scons: building terminated because of errors.
```

When i force it to run all tests via `scons -k` i get more errors like this, some c++ related and 2 tests are marked as failed.
```
[FAIL]   Pattern not found in build\options\options.pb.h: \s+PB_MSG\(104,[0-9]*,Message4\) \\

scons: *** [build\options\options_h.matched] Error 1
match_files(["build\options\proto3_options.pb.matched"], ["build\options\proto3_options.pb.h", "build\options\proto3_options.expected"])

[FAIL]   Program build\regression\issue_1061\test.exe returned 1
scons: *** [build\regression\issue_1061\test.output] Error 1
match_files(["build\regression\issue_125\extensionbug.pb.matched"], ["build\regression\issue_125\extensionbug.pb.h", "build\regression\issue_125\extensionbug.expected"])
```

I just assumed this might be the state of things and went with it, checked the added test case ran and produced the desired output. Sorry for the confusion, propably should have asked for instructions first, just did not want to bother you with this.

You are correct about your second point. If we define package it should be handled by the existing logic either in the self.replacement_prefix case or at the file_options.package condition.

I will update and test again as soon as i get to it.

PetteriAimonen:
Looks good now, thanks!

PR #1135, opened 2026-02-01
Fix toolchains list in extra/bazel/nanopb_cc_proto_library.bzl

gonzojive:
Update the toolchain list in extra/bazel/nanopb_cc_proto_library.bzl to match the example provided by
https://rules-proto-grpc.com/en/latest/custom_plugins.html.

This PR is intended to be merged after https://github.com/nanopb/nanopb/pull/1134.

### DIFF
--- a/extra/bazel/nanopb_cc_proto_library.bzl
+++ b/extra/bazel/nanopb_cc_proto_library.bzl
@@ -8,6 +8,7 @@ load(
     "ProtoPluginInfo",
     "proto_compile",
     "proto_compile_attrs",
+    "proto_compile_toolchains"
 )
 load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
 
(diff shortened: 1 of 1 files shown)


schultetwin:
I see 3 changes in this PR:

1. 🟢 (From #1134) - Use `rules_cc` for `cc_library`. 
This is necessary for Bazel 9, as rulesets for building C++/C files have moved out of core Bazel and into rules libraries. This makes sense, and should be backward compatible.
2. 🟢 (From #1134) - Use Buildifier to lint `.bzl` and `BUILD` files. 
These are just formatting changes and have no real impact, but a nice clean-up. Would be nice to put this in CI somehow. This has no compatibility issues.
3. 🟢  Don't reference a `toolchain` directly, but instead use `proto_compile_toolchains` from `rules_proto_grpc`. 
This seems fine to me. @gonzojive, I'd love some context on why this is necessary? I assume you're using a newer version of `rules_proto_grpc` then `nanopb` is using, and this allows `nanopb` to work with the newer version? But this should be backward compatible with the version of `rules_proto_grpc` that `nanopb` is using.

Overall, changes LGTM

gonzojive:
This was necessary to use [an aspect rule that copies generated files into the source tree.](https://registry.bazel.build/docs/bazel_lib#function-write_source_files), I believe.

PR #1146, opened 2026-03-03
fix(platformio): handle subdirectories when preserve_directory_hierarchy=true

jarkad:
This fixes #1145, which I submitted without testing. My apologies for that.

When proto-path is not specified, nanopb resolves path to the .proto file
relative to its working directory. This commit makes nanopb run from the project
root (set by `project_dir` option), and provides full relative path to the
.proto files.

---

#1145 was reverted in 99c830db6238892cea1d7957f7196c20bc5bbce6, probably because of this problem, so I'm including it again.

Again, sorry for the inconvenience!


PetteriAimonen:
Ah, sorry about the revert, it actually wasn't intentional. I just messed up when merging my local changes.

PR #1170, opened 2026-06-29
Update actions and fix ci for forks

jsoref:
If you create a fork as github suggests and don't include all branches + tags, then CI fails miserably, see: https://github.com/check-spelling-sandbox/nanopb/actions/runs/28336774704
While it's in the process of failing, GitHub complains the the actions are using old versions of node.

- This PR fixes the main failure by defaulting to `nanopb-0.0.0` when `git describe --always` doesn't find a tag
- This PR fixes the warnings by upgrading the actions to the current versions

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -19,10 +19,10 @@ jobs:
         os: [ubuntu-latest, macos-latest]
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
 
       - name: Mount bazel cache
-        uses: actions/cache@v4
+        uses: actions/cache@v6
         with:
           path: "/home/runner/.cache/bazel"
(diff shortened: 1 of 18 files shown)


jsoref:
19352323318fbd02e7f21723495d9c3f82b939ff would fix https://github.com/nanopb/nanopb/actions/runs/28343375402/job/84354443547?pr=1170, see https://github.com/check-spelling-sandbox/nanopb/actions/runs/28466068600?pr=1

It's an unrelated problem -- it happens to PRs within a single repository: https://github.com/check-spelling-sandbox/nanopb/actions/runs/28464973529?pr=1

Happy to add it here, or make it a distinct PR.

PetteriAimonen:
If you can add that fix, I would be very glad!

PetteriAimonen:
Thanks a lot!

PR #1171, opened 2026-06-29
Encoding optimisation for small, buffer-only submessages

btRooke:
## Overview

If a submessage is < 128 bytes, we can skip the sizing step by reserving a byte for length and writing it in after the submessage has been encoded.

If it turns out to be larger, we can reset the stream and encode as we usually would.

In the ideal case, this drops the number of `pb_encode_submessage()` calls for an $n$-nested message from $n^2$ to $n$.

You get no where near the speedup you would vs manually encoding a message but it's probably worth getting in as messages are very frequently (like most in the test suite) < 128 bytes.

Haven't searched too much so apologies if this has been considered before and rejected!

## Testing

Added a temporary (prints time to encode to `stderr`) timing test to demonstrate the difference. I used the `.proto` files from the `deep_structure` test to demonstrate an extremely bad case + improvement (~9x speedup).

Also added a test with a submessage >= 128 bytes as I couldn't find one and also added this to the buffer-only tests.

Sorry for any abuse of the test framework, I'm not an expert!

### DIFF
--- a/pb_encode.c
+++ b/pb_encode.c
@@ -731,6 +731,69 @@ bool checkreturn pb_encode_string(pb_ostream_t *stream, const pb_byte_t *buffer,
 }
 
 bool checkreturn pb_encode_submessage(pb_ostream_t *stream, const pb_msgdesc_t *fields, const void *src_struct)
+#ifdef PB_BUFFER_ONLY
+{
+    /*
+     * If we're in buffer-only mode we can skip the submessage sizing step by
+     * just reserving a byte on the stream, encoding the submessage, then
+     * writing the length in afterwards.
+     *
+     * We can't do this in stream mode as we can't go back edit the reserved
+     * byte.
+     *
+     * This only works when the submessage is <= 127 bytes (max single-byte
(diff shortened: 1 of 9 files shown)


PetteriAimonen:
This is actually already implemented in the dev1.0 branch, I'm just having a hard time finding motivation & time to finish the branch up enough for merging. It reserves one byte, but if the message is longer it uses `memmove()` to make more space.

https://github.com/nanopb/nanopb/blob/a90d58ccf694883fe5b0e1613b7c8c3928337db6/src/pb_encode.c#L646

I think your implementation for 0.4.x is elegant in its simplicity. But I think I'll keep 0.4.x maintenance as mostly bugfix-only, at least if I manage to get 1.0 out soon enough..

btRooke:
Great minds think alike...

Makes sense regarding next 0.4 release, looking forward to seeing the improvement in 1.0.

And good call about `memmove()`, could definitely swap the second `pb_encode()` for that in mine.

Happy to close this PR.

PR #1172, opened 2026-07-02
feat(bazel): Upgrade to Bazel 8 + proto support to @protobuf

schultetwin:
## Problem

nanopb's Bazel build depended on the deprecated rules_proto and rules_proto_grpc for the cc_nanopb_proto_library implementation, and on an outdated protobuf 29.0. Upstream is consolidating all Bazel proto support in the protobuf module itself.

## Solution

* Update protobuf to 35.1 and drop the rules_proto and rules_proto_grpc dependencies. protobuf 35.x requires Bazel 8, so .bazelversion is now 8.7.0.
* Reimplement cc_nanopb_proto_library on the proto_common API from @protobuf. The rule interface is unchanged.
* Replace the rules_proto_grpc proto_plugin target with a proto_lang_toolchain, supporting both legacy toolchain lookup and --incompatible_enable_proto_toolchain_resolution (the Bazel 9 default).
* Remove the WORKSPACE-era helper macros; WORKSPACE builds are no longer supported by protobuf 35.x.
* CI now tests against the latest Bazel 8 and Bazel 9 releases.

## Discussion

* *Bazel 7*: This PR drops support for Bazel 7. Bazel 7 goes [end of life](https://bazel.build/release) in December this year (2026). It's unfortunate to end support for Bazel 7 earlier than end of life. However, @protobuf has already removed support for Bazel 7, so I don't see a large regression here.
* *WORKSPACE Removal*: This PR drops support for Bazel `WORKSPACE` files. With Bazel 8, the WORKSPACE file is disabled by default. Users should be able to use the bazelmod variant.
* *CI*: This PR adds CI for Bazel 8 and Bazel 9. I think this is okay, just want to call it out here.


PetteriAimonen:
@Sayter99 @tpudlik Do you have time to take a look at this?

hiSandog:
The Bazel 8 upgrade is a good place to add a small compatibility matrix in CI. At minimum, one test should build a proto with nanopb options using the new `@protobuf` path and another should verify an existing Bazel 7-style workspace either still works or fails with a documented migration message. That would make it clear whether this PR is a hard migration or a bridge release.

schultetwin:
> At minimum, one test should build a proto with nanopb options using the new @protobuf path

All of the tests now build with @protobuf? Not following what you're asking for here, but happy to add a new test if I'm missing something!

>  another should verify an existing Bazel 7-style workspace either still works or fails with a documented migration message. That would make it clear whether this PR is a hard migration or a bridge release.

Yes, this is a hard break. I'm not sure we need a CI test that tests to see if we fail on Bazel 7, but I did add explicit failure messages on the old `load` paths that people would use in Bazel 7.

extra/bazel/protoc-gen-nanopb-bazel.py
Sayter99, line 1: does the built-in generator work for major platforms? if it doesn't take much time and can work for more environment by default I think we might not need this custom stuff. Could you provide more info on why we need this?

extra/bazel/BUILD.bazel
Sayter99, line 1: could you elaborate more on why it is required?

extra/bazel/protoc-gen-nanopb-bazel.py
schultetwin: This file (along with the `BUILD` file in this folder) are here mainly as a small tweak to how Bazel launches the generator. It should not change the environments that the generator can run in.

Previously, in the `BUILD` file in the root of this repo, we had:

```bazel
copy_file(
    name = "protoc-gen-nanopb.py",
    src = "generator/protoc-gen-nanopb",
    out = "generator/protoc-gen-nanopb.py",
    allow_symlink = True,
)

py_binary(
    name = "protoc-gen-nanopb",
    srcs = glob([
        "generator/**/*.py",
    ]) + [
        ":protoc-gen-nanopb.py",
    ],
    data = glob([
        "generator/**/*.proto",
    ]),
    env = {
        "NANOPB_PB2_NO_REBUILD": "1",
    },
    imports = [
        "generator",
    ],
    deps = [
        ":nanopb_py_proto",
    ],
)
```

We had to rework this for the following reasons:

The `env =` attribute on the `py_binary` was actually dead code. `env` on executable rules only applies to `bazel run` and `bazel test` — it does not apply when the binary is spawned as a tool inside a build action. The `NANOPB_PB2_NO_REBUILD=1` that mattered during proto compilation was injected by something else entirely: `rules_proto_grpc`'s `proto_plugin(env = ...)`, whose compile rule set env on the protoc action it created.

The migration removed that injection point. protobuf's `proto_common.compile()` has no env parameter (it runs protoc with use_default_shell_env = True), so with `rules_proto_grpc` gone, there was nowhere left in the rule layer to set the variable. And the variable genuinely matters: `nanopb_generator.py` calls `load_nanopb_pb2()` at import time — without the env var it tries to rebuild `nanopb_pb2.py` by invoking whatever `protoc` it can find, which inside the action sandbox either fails or nondeterministically uses a wrong-version system `protoc`. With it, the generator does a plain `import nanopb_pb2`, served hermetically by the `py_proto_library` dependency. Because the check runs at import, before option parsing, a plugin flag couldn't replace it either.

This spawns the creation of `extra/bazel/protoc-gen-nanopb-bazel.py`, a Bazel-only entry point that sets `os.environ["NANOPB_PB2_NO_REBUILD"] = "1"` before importing the generator. Setting it in the plugin process itself is the only placement that survives being spawned by `protoc` regardless of what creates the action. And since that wrapper is a real `.py` file, it serves directly as the `py_binary`'s main — making the `copy_file` redundant rather than just moved.

extra/bazel/BUILD.bazel
schultetwin: The `exports_fiels` is required so that the `BUILD` file at the root of the repo can reference `protoc-gen-nanopb-bazel.py` directly.

The `toolchain_type` registers a toolchain type for nanopb as is described here: https://docs.google.com/document/d/1CE6wJHNfKbUPBr7-mmk_0Yo3a4TaqcTPE0OWNuQkhPs/edit?tab=t.0#heading=h.5mcn15i0e1ch.

This will allow users to override and inject their own nanopb customizations.

Sayter99 (approved):

PetteriAimonen:
Thanks both!

PR #1173, opened 2026-07-05
Improve diagnostics for invalid options patterns

### LINKED ISSUE #954: Improved diagnostics for stray colon in .options files
Compare the two lines below:

```
edv.GatewayMessage  type: FT_POINTER
edv.GatewayMessage: type: FT_POINTER
```

Side by side it is relatively visible, but can easily get lost when editing the file. The diagnostic emitted in this case is also quite enigmatic:

```
Following patterns in proto/engine/diagnostics/v1/gateway.options did not match any fields: edv.GatewayMessage:
```

It would be amazing if we could a better diagnostic.

Old-Ding:
﻿## Summary
- validate .options field patterns as soon as they are read
- report the invalid character and a whitespace hint for stray ':' separators
- add a regression test for issue #954, including a valid .proto file pattern

## Tests
- python -m py_compile generator\nanopb_generator.py tests\regression\issue_954\test_options_diagnostics.py
- direct tests\regression\issue_954\test_options_diagnostics.py run in a temporary Python 3.12 venv with generated nanopb_pb2.py
- SCons single target for regression/issue_954 run in the same temporary venv
- git diff --check

Fixes #954

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -2390,6 +2390,24 @@ class ProtoFile:
 
 from fnmatch import fnmatchcase
 
+def validate_options_namemask(namemask, filename, line_number):
+    '''Verify that an options pattern contains only protobuf path and fnmatch characters.'''
+    # File options match .proto paths, which can contain punctuation that is
+    # not valid in field names.
+    if '/' in namemask or namemask.endswith('.proto'):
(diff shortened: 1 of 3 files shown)


PetteriAimonen:
(The Mac build CI break is probably something unrelated)

PetteriAimonen:
Thanks!

PR #1174, opened 2026-07-05
cmake: require 3.5 for legacy examples

### LINKED ISSUE #1128: QA Notice: Compatibility with CMake < 3.5 has been removed from CMake 4
```
 * QA Notice: Compatibility with CMake < 3.5 has been removed from CMake 4,
 * dev-libs/nanopb will fail to build w/o a fix.
 * See also tracker bug #951350; check existing or file a new bug for this package.
 * Please also take it upstream.
 * 
 * The following files are causing errors:
 *   examples/cmake_simple/CMakeLists.txt:2.8
 *   examples/cmake_relpath/CMakeLists.txt:2.8
 *   conan-wrapper/CMakeLists.txt:2.8.12
 *   build-tests/legacy_cmake_simple/CMakeLists.txt:2.8
 *   build-tests/legacy_cmake_relpath/CMakeLists.txt:2.8
 *   build-tests/cmake_with_components/CMakeLists.txt:2.8
 * 
 * Compatibility with CMake < 3.10 will be removed in a future release.
 * See also tracker bug #964405; check existing or file a new bug for this package.
 * If not fixed in upstream's code repository, we should make sure they are aware.
 * 
 * CMake 4 detected; building with -DCMAKE_POLICY_VERSION_MINIMUM=3.5
 * This is merely a workaround to avoid CMake Error and *not* a permanent fix;
 * there may be new build or runtime bugs as a result.
 * 
 * An upstreamable patch should take any resulting CMake policy changes
 * into account. See also:
 *   https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html

```

Old-Ding:
﻿## Summary
- raise the remaining example and build-test `cmake_minimum_required` declarations from 2.8/2.8.12 to 3.5
- keep the existing CMake logic unchanged while avoiding CMake 4's removal of compatibility with versions older than 3.5

Fixes #1128

## Testing
- `git diff --check HEAD~1..HEAD`
- scanned CMake files for remaining minimum versions below 3.5
- `git ls-files --eol` for the changed CMakeLists files
- Not run: CMake configure/build, because `cmake` is not installed on this Windows host.

### DIFF
--- a/build-tests/cmake_with_components/CMakeLists.txt
+++ b/build-tests/cmake_with_components/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.5)
 project(NANOPB_CMAKE_SIMPLE C CXX)
 
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../extra)
--- a/build-tests/legacy_cmake_relpath/CMakeLists.txt
+++ b/build-tests/legacy_cmake_relpath/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.5)
 project(NANOPB_CMAKE_SIMPLE C)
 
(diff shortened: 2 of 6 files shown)


PetteriAimonen:
Thanks!

Issue #1031, opened 2024-11-08
Can't set different max_count values for different sub-messages of the same type


philbinj:
For example, for this proto:
```
message FloatArray {
  repeated float xs = 1;
}
message Msg {
  FloatArray a = 1;
  FloatArray b = 2;
}
```
How can I set different `max_count` values for `a` and `b`? The following options file doesn't seem to work:
```
Msg.a.xs max_count:50
Msg.b.xs max_count:100
```

PetteriAimonen:
They both use the same submessage type, so it is not possible by just using nanopb options. Because the same C struct is used for both, they cannot have different sizes.

If the arrays were directly in the main message (instead of wrapped in a submessage), this would be possible. Another way would be to have separate submessage types in `.proto`.

You can do this with `callback_datatype` and a custom callback function, but that quickly gets pretty complex.

PR #1175, opened 2026-07-05
Fix comment parsing in options string literals

Old-Ding:
## Summary
- replace the order-dependent comment regex passes with one string-aware regex
- preserve //, #, and /* */ markers when they are part of quoted option values
- keep line numbers stable when removing multiline block comments
- extend the existing options diagnostics regression to cover all three comment markers in strings
- run the diagnostics regression with the active SCons Python interpreter so binary package tests use the environment that has protobuf installed

## Duplicate check
- No open PR found for `read_options_file`, `options comment string callback_datatype`, or `callback_datatype`.
- Open issue search for `callback_datatype` only found #1031, which is unrelated to comment parsing.

## Tests
- `./.venv/Scripts/python.exe tests/regression/issue_954/test_options_diagnostics.py`
- From `tests`: `../.venv/Scripts/python.exe -m SCons build/regression/issue_954/test_options_diagnostics.output`
- From `tests`, with `.venv/Scripts` prepended to `PATH`: `../.venv/Scripts/python.exe -m SCons build/regression/issue_145/comments.pb.h`
- 51-case differential check against the previous scanner: 45 repository `.options` files plus quoted markers, escapes, multiline comments, and unterminated input; 0 mismatches
- `./.venv/Scripts/python.exe -m py_compile generator/nanopb_generator.py tests/regression/issue_954/test_options_diagnostics.py`
- `git diff --check`
- `git ls-files --eol generator/nanopb_generator.py tests/regression/issue_954/test_options_diagnostics.py tests/regression/issue_954/SConscript`

### DIFF
--- a/generator/nanopb_generator.py
+++ b/generator/nanopb_generator.py
@@ -2408,15 +2408,27 @@ def validate_options_namemask(namemask, filename, line_number):
                          (invalid.group(0), namemask, hint))
         sys.exit(1)
 
+def strip_options_comments(data):
+    '''Remove comments from .options data without touching quoted strings.'''
+    string_or_comment = re.compile(
+        r'''(?P<string>"(?:\\.|[^"\\\r\n])*(?:"|$)|'(?:\\.|[^'\\\r\n])*(?:'|$))'''
+        r'''|(?P<comment>/\*.*?(?:\*/|\Z)|//[^\r\n]*|#[^\r\n]*)''',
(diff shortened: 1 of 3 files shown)


PetteriAimonen:
The parser seems somewhat complex. I wonder if there might be some Python built-in or other simpler approach to the same.

PetteriAimonen:
Thanks!

PR #1177, opened 2026-07-14
Regenerate PlatformIO output after options removal

94xhn:
## Problem

`platformio_generator.py` records an MD5 marker when a matching `.options` file exists. If that options file is later deleted while the `.proto` file stays unchanged, the old marker is ignored and generation is skipped. The existing `.pb.c`/`.pb.h` output therefore keeps options that are no longer present.

## Fix

- keep the expected options-marker path even when the `.options` file is absent
- treat a missing options file with a stale marker as a cache miss
- remove that marker only after regeneration succeeds, so a failed run retries
- add regression coverage for create, modify, delete, and failed-regeneration states

## Verification

- The new two-test regression suite fails 2/2 against `master`.
- Windows Python 3.10: 2/2 pass.
- WSL Python 3.12: 2/2 pass.
- `py_compile`, Ruff, formatting, and `git diff --check` pass for the new test and changed files.
- Nanopb core C sources compile with GCC 13, `-Wall -Wextra -Werror`, ASan, and UBSan.

The full PlatformIO package/example build was not run locally because `pio` is not installed. The existing PlatformIO workflow still runs that build and now runs this focused regression first. There is no firmware Flash or RAM impact; this changes generator cache handling only.

### DIFF
--- a/.github/workflows/platformio_tests.yml
+++ b/.github/workflows/platformio_tests.yml
@@ -34,6 +34,9 @@ jobs:
           python3 -m venv venv
           venv/bin/pip3 install -U platformio
 
+      - name: Test PlatformIO generator cache
+        run: venv/bin/python3 -m unittest nanopb/build-tests/platformio/test_platformio_generator.py
+
       - name: Build PlatformIO package
         run: |
           source venv/bin/activate
--- a/build-tests/platformio/test_platformio_generator.py
+++ b/build-tests/platformio/test_platformio_generator.py
@@ -1,0 +1,160 @@
+import hashlib
+from pathlib import Path
+import runpy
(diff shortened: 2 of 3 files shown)


PetteriAimonen:
`test_platformio_generator.py` shouldn't be under the generator folder. It can go in e.g. `build-tests/platformio`.

94xhn:
Moved the regression test to build-tests/platformio and updated the workflow path in 06eff88. I reran the test from the same parent-directory layout used by Actions on Windows Python 3.10 and WSL Python 3.12; both pass. Thanks.

PetteriAimonen:
Thanks.

PR #1183, opened 2026-09-01
Fix missing _proto directory in Windows package

### LINKED ISSUE #1182: 0.4.9.2 Windows binary package missing generator-bin/grpc_tools/_proto directory
Using the nanopb-0.4.9.2-windows-x86.zip binary package, nanopb_generator.exe produces the following warning for each .proto file:

nanopb-windows-x86\generator-bin\grpc_tools\_proto: warning: directory does not exist.

The ZIP contains:

"generator-bin\grpc_tools\google\"

but does not contain:

"generator-bin\grpc_tools\\_proto\"

Creating an empty "generator-bin\grpc_tools\\_proto" directory is sufficient to remove the warnings, and generation then completes normally.

This appears to be a minor packaging issue in the 0.4.9.2 Windows ZIP rather than a generator failure?  Apologies in advance if this is user error XD

Huoyanlifusu:
Fixes #1182.

Create `generator-bin/grpc_tools` before copying `_proto`, so `cp` keeps the `_proto` directory in the Windows package instead of copying only its contents.

Tested with the binary package workflow: https://github.com/Huoyanlifusu/nanopb/actions/runs/33457513112. All three jobs passed, including Windows 2022.

### DIFF
--- a/tools/make_windows_package.sh
+++ b/tools/make_windows_package.sh
@@ -29,6 +29,7 @@ mv $DEST/generator/dist/nanopb_generator $DEST/generator-bin
 cp $DEST/generator/dist/protoc/protoc.exe $DEST/generator-bin
 
 # Include Google's descriptor.proto and nanopb.proto
+mkdir -p $DEST/generator-bin/grpc_tools/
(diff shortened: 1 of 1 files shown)


PetteriAimonen:
Thanks!

Maybe I should update the 0.4.9.2 Windows binary package to include the fix.

PR #1184, opened 2026-09-07
Never allocate less than sizeof(pb_bytes_array_t) for a bytes array

### LINKED ISSUE #1180: PB_BYTES_ARRAY_T_ALLOCSIZE can sometimes return a size < sizeof(pb_bytes_array_t), causing compilation errors with -Warray-bounds
https://github.com/nanopb/nanopb/blob/d21fa5084287ab67da2f166f4def045bedcb535e/pb.h#L403

`PB_BYTES_ARRAY_T_ALLOCSIZE(n)` is defined as `((size_t)n + offsetof(pb_bytes_array_t, bytes))`
`pb_bytes_array_t` is defined as:
```
struct pb_bytes_array_s {
    pb_size_t size;
    pb_byte_t bytes[1];
};
typedef struct pb_bytes_array_s pb_bytes_array_t;
```

If we have set `PB_FIELD_32BIT`, `pb_size_t` is a uint32. Hence `sizeof(pb_bytes_array_t)` must be at least 5, likely 8 on systems that pad for alignment.
When If we were to allocate a region of `PB_BYTES_ARRAY_T_ALLOCSIZE(0)`, we would allocate 4 bytes, less than `sizeof(pb_bytes_array_t)`. On systems that pad to 4/8bytes, allocations for arrays of 1, 2, 3 bytes also allocate less than the size of the structure.

In practice most malloc implementations would pad the allocation to 4/8bytes, and the user should respect the `size` attribute, not writing outside the allocated memory region. However the macro definition as stands can cause compilation errors in gcc when array-bounds warnings are enabled (-Warray-bounds).

swimmingbrain:
`PB_BYTES_ARRAY_T_ALLOCSIZE(n)` is `offsetof(bytes) + n`, which for a small `n` is less than `sizeof(pb_bytes_array_t)` as soon as `pb_size_t` gets padding after it, with `PB_FIELD_32BIT` for example. gcc 13 and newer see a `pb_bytes_array_t` pointer into a smaller `malloc` and refuse it with `-Warray-bounds`, as in #1180.

The macro now rounds up to `sizeof(pb_bytes_array_t)`. The two fuzz test helpers used `ALLOCSIZE(0)` to mean the header in front of the bytes, they say `offsetof` now so they keep the same value. Checked with gcc 14: the reproducer from the issue fails on master with `-DPB_FIELD_32BIT -Warray-bounds -Werror` and compiles and runs with this change, and the scons test suite passes.

Closes #1180


PetteriAimonen:
Thanks! I'll wait for the CI to pass before merging.

Do you agree with my assessment that there is no security impact from this bug?
It's a bit borderline whether casting the too small allocation into a `pb_bytes_array_t*` is undefined behavior, even though there is never more than `size` bytes written into it.
Theoretically there could be some runtime bounds checker that could get tripped and cause a denial-of-service when empty bytes value is received.

swimmingbrain:
Thanks for merging!

I agree, no security impact. Every write through that pointer is either `size` or `bytes[0..size-1]`, and both stay inside the block that was actually allocated, so nothing lands out of bounds at runtime. What is wrong is only the type: the object is smaller than `sizeof(pb_bytes_array_t)`, so a tool reasoning from the declared type rather than from the accesses can object, which is exactly what gcc does with -Warray-bounds.

A runtime checker doing whole-object bounds could abort on the same grounds, and an empty bytes field is reachable from input, so the worst case I can construct is a crash in a build that has such checking switched on. That is availability only, and it needs the checker enabled in production, which is not how people ship. So I would call it a portability fix rather than a security one.

Issue #543, opened 2020-06-21
Self-tests failure on some architectures


gcsideal:
Hi,
nanopb packaged for Debian and builds on all architectures it supports. But [self-tests are failing](https://buildd.debian.org/status/package.php?p=nanopb&suite=sid) on some architectures as:
run_test(["build/float_double_conversion/float_double_conversion.output"], ["build/float_double_conversion/float_double_conversion"])
Command line: ['build/float_double_conversion/float_double_conversion']
FAILED:[22;39m build/float_double_conversion/float_double_conversion.c:62 memcmp(&msg.value, &expected_float, sizeof(float)) == 0
FAILED:[22;39m build/float_double_conversion/float_double_conversion.c:76 memcmp(&msg.value, &expected_double, sizeof(double)) == 0
FAILED:[22;39m build/float_double_conversion/float_double_conversion.c:62 memcmp(&msg.value, &expected_float, sizeof(float)) == 0
FAILED:[22;39m build/float_double_conversion/float_double_conversion.c:76 memcmp(&msg.value, &expected_double, sizeof(double)) == 0
[FAIL][0m   Program build/float_double_conversion/float_double_conversion returned 1

I don't know if it's a coincidence, but the failing architectures are big-endian and the comparison may fail due to the byte order.

PetteriAimonen:
Hmm, it could be endianess issue in the PB_CONVERT_DOUBLE_FLOAT 
functionality. I *think* it should work for big-endian also, but
it would certainly break for mixed-endian architectures where FPU 
endianess would differ from integer endianess.

It's not an option that would be used on most architectures, it is 
mostly for AVR which has no 64-bit double type. Currently the test runs 
on all architectures but it certainly doesn't have to.

I'll have to look up if there would be some C compiler define or similar 
I could use to detect this.

Also, thanks for the packaging. I took a brief look and they look 
reasonable :)

PetteriAimonen:
Turns out it was a portability bug in the PB_CONVERT_DOUBLE_FLOAT 
implementation, which failed to handle NaN representations with bottom
bits set.

The test cases succeed for mips and mipsel for me now.

gcsideal:
> Turns out it was a portability bug in the PB_CONVERT_DOUBLE_FLOAT implementation, which failed to handle NaN representations with bottom bits set. The test cases succeed for mips and mipsel for me now.

Just tested current 'master' on one of our mipsel machines. Unable to confirm, fails the same way.
[...]

PR #1185, opened 2026-09-08
Preserve NaN values in double-to-float decoding

vpfkfl753:
With `PB_CONVERT_DOUBLE_FLOAT`, a double NaN whose payload fits entirely in the discarded 29 bits is decoded as infinity. For example, `0x7ff0000000000001` becomes `0x7f800000`.

Keep a nonzero float mantissa in this case. Adds raw-wire tests for positive and negative NaNs around the mantissa cutoff; four cases fail before the fix.

The float/double conversion tests and the SCons suite pass with `NOVALGRIND=yes NOUNSAFE=yes` on Linux.

Related to #543.


PetteriAimonen:
Thanks!

PR #1186, opened 2026-09-08
Fix runtime sources marked as generated by nanopb_generate_cpp

### LINKED ISSUE #1104: FindNanopb.cmake pb_common.c, pb_decode.c and pb_encode.c marked for cleanup during ninja clean
When using the "traditional variables" version of NANOPB_GENERATE_CPP, the non-generated source files in NanoPB get clobbered on a ninja clean.  My workaround was to remove them from the NANOPB_SRCS list after find_package() statement and then add them back manually to our targets using target_sources() after our NANOPB_GENERATE_CPP statements.

vpfkfl753:
Fixes #1104.

When `NANOPB_SRCS` and `NANOPB_HDRS` are reused as output variables, `nanopb_generate_cpp()` marks the runtime sources as `GENERATED` too. In a copy of `legacy_cmake_simple`, `ninja clean` then deleted `pb_common.c`, `pb_decode.c` and `pb_encode.c`, and the next build failed.

Remove the blanket property assignment. [`add_custom_command(OUTPUT ...)`](https://cmake.org/cmake/help/v3.5/command/add_custom_command.html) already marks the generated files automatically. A separate `legacy_cmake_generated_sources` build test checks this case; the existing usage example is unchanged. No changes are required in users' CMake files.

Tested on Linux with CMake 3.28.3 and GCC 13.3: the isolated regression fails before the fix; build, clean, rebuild and execution pass with Ninja and Unix Makefiles after it. The restored legacy example also builds and runs. The two CMake examples and three original CMake build-test projects passed in the initial validation.


PetteriAimonen:
Can you clarify whether the change to the build_tests case is a required change for users (breaking change), or just a regression test for this bug?

If it is just a regression test, I think putting it in the file like this runs the risk that someone accidentally copies the weird rule to their own project file.

vpfkfl753:
It is only a regression test; users do not need to change their CMake files. I restored `legacy_cmake_simple` and moved the check to a separate `legacy_cmake_generated_sources` build test, included in the build-test workflow.

The isolated test fails with the old module. Build, clean, rebuild and execution pass with Ninja and Unix Makefiles after the fix; the original example also builds and runs.

PetteriAimonen:
Thanks!

PR #1188, opened 2026-09-16
Make enum_intsize work outside C23 and C++11

daniel22-jung-ivis:
## Problem

`enum_intsize` generates `enum Foo : uint8_t`. That syntax is standard only in
C++11 and C23, so whether the generated header compiles depends on the
toolchain and on how strictly the build is configured:

| situation | result today |
|---|---|
| C++11 / C23 build | works |
| GCC or clang in an older C mode, default flags | accepted as an extension, silently |
| the same, with `-pedantic-errors` or `-Wpedantic -Werror` | **build fails** |
| a toolchain without such an extension | **build fails** |

So this is not simply about age. A project can sit on C99 and never notice,
until it turns on conformance checking, or moves to a compiler that does not
offer the extension. Then the header stops compiling:

```
enum_intsize.pb.h:13:14: error: ISO C does not support specifying 'enum'
                                underlying types before C23 [-Wpedantic]
   13 | typedef enum _IntSizeInt8: uint8_t {
```

Relying on an extension is also not free even while it works, since the
generated header is then not the standard C it claims to be.

This matters most to the projects the option is aimed at. Embedded code is
often built in strict conformance mode because a coding standard requires it,
or is stuck on a vendor toolchain whose version is fixed by certification, and
those are the same projects that count bytes of RAM.

Building the current `tests/enum_sizes/enum_intsize.proto` across the
compilers and standards available on one machine, with `-pedantic-errors`:

| | before | after |
|---|---|---|
| combinations that build | 15 / 30 | **29 / 30** |

The one remaining failure is `clang -std=c89 -pedantic-errors`, which stops on
`bool` in `pb.h` and has nothing to do with this option.

## How this differs from the `#ifdef` that 0.4.9 removed

An `#if` around the enum definition should raise an eyebrow, so here is why
this is not the construct that was taken out in b34a649.

0.4.7 generated this:

```c
typedef enum _Foo
#ifdef __cplusplus
 : uint8_t
#endif
{ ... } Foo;
```

The C compiler saw a plain enum and dropped the size request on the floor,
while the C++ compiler honoured it. Same header, two different layouts:

| | 0.4.7, removed in b34a649 | this PR |
|---|---|---|
| branch selected by | presence of `__cplusplus` | language standard version |
| C result | default size enum, **4 bytes** | `typedef uint8_t`, **1 byte** |
| C++ result | `: uint8_t`, 1 byte | `: uint8_t`, 1 byte |
| the two branches | **disagree, silent ABI break** | **agree** |

The difference is that the fallback here still delivers the requested size,
using different syntax, instead of ignoring the option. b34a649 turned that
silent breakage into a compiler error on purpose, and this PR keeps that
property: there is no configuration in which the layout can differ.

Using a compiler extension instead (for example GCC's `mode` attribute) also
gives a matching layout, but only on compilers that have one. It leaves IAR,
ARM Compiler 5 and similar toolchains failing, and it puts a table of compiler
specific knowledge into `pb.h` that has to be maintained.

## Approach

The generated header carries both forms and lets the compiler choose, using
only the version macros the language standards themselves mandate:

```c
#if (defined(__cplusplus) && __cplusplus >= 201103L) || \
    (!defined(__cplusplus) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L)
typedef enum _DeviceStatus2: uint8_t {
    DeviceStatus2_OFFLINE = 0,
    DeviceStatus2_ONLINE = 1,
    DeviceStatus2_ERROR = 2
} DeviceStatus2;
#else
typedef uint8_t DeviceStatus2;
#define DeviceStatus2_OFFLINE ((DeviceStatus2)0)
#define DeviceStatus2_ONLINE ((DeviceStatus2)1)
#define DeviceStatus2_ERROR ((DeviceStatus2)2)
#endif
```

`pb.h` is not touched, and there is no compiler detection anywhere. Any C or
C++ compiler resolves this on its own.

Defines rather than a second enum are used for the values because C requires
enumeration constants to fit in `int`, which does not hold for 16-bit `int`
targets such as AVR.

## Binary compatibility

This is the part that matters, so it is covered by a test rather than an
argument. Both forms are the same unsigned integer type of the same width, so
the choice of branch does not change any layout. `tests/enum_sizes` now builds
one translation unit as C and one as C++, from the same generated header, and
links them:

```
gcc -ansi     -c enum_intsize_mixed_c.c      # takes the typedef branch
g++ -std=c++11 -c enum_intsize_mixed_cpp.cpp # takes the enum branch
```

The C side fills a message and passes `sizeof`/`offsetof` to the C++ side,
which compares them against its own and checks the field values. The test fails
if the two branches ever drift apart.

## What does not change

Enums that do not use `enum_intsize` generate byte for byte the same output as
before. Verified by generating the existing test protos with both the old and
the new generator and diffing the results:

```
$ diff -r old/ new/
(no differences, 10 files)
```

## Values that do not fit are now rejected

The `enum : uint8_t` form makes the compiler reject an out of range enumerator,
but the fallback would quietly truncate it in the `((Name)value)` cast. To keep
the two forms equivalent, the generator now refuses such values:

```
Enum 'Level' value TOO_HIGH = 300 does not fit in uint8_t selected by
enum_intsize (valid range is 0..255). Use a larger enum_intsize, or remove
the option to keep the default 'int' representation.
```

Negative values get their own message, since a larger size would not help:
every `enum_intsize` maps to an unsigned type.

## Tests

`tests/enum_sizes` gains three tests, alongside the existing C++11 one:

- the proto built as C with `-ansi -Werror -pedantic-errors`. The flag is
  deliberate: without it GCC and clang accept the old syntax as an extension,
  and the test would pass even against the unfixed generator
- the C and C++ translation units linked together, described above
- a generator test for the range check, including the boundary values

## How to reproduce

```bash
cd tests && scons build/enum_sizes
```

## Known limitations

- In the fallback branch the type is a plain integer, so `-Wswitch` and similar
  warnings are not available there. C++11 and C23 builds are unaffected.
- Enums with negative values cannot use the option, as the generated type is
  unsigned. They were already broken before this change, but silently.


PetteriAimonen:
Why do you want to specify the intsize for enum if your compiler doesn't support it?
And please write your answer by human, these AI-generated summaries are too verbose.

In initial view the generated code looks way too complex and ugly to merge, for such a niche usecase.

daniel22-jung-ivis:
Thank you for your feedbacks.

The same .proto is sometimes used in both environments, one where the compiler supports it and one where it doesn't. 
That is why I tried to improve this.

In that case, could I ask what approach you would consider right?

There is no problem if you decide not to merge it.

Thanks.

PetteriAimonen:
I'm not sure there exists any method that I would consider worth the maintenance burden, as this is quite rare use case.

As for sharing `.proto` files, I would consider just having separate `.options` file for the other environment to unset the enum size. Though practically, it's rarely worth the time to optimize a few bytes from the size of an `enum` and you could just keep the default enum size.

daniel22-jung-ivis:
Thanks for the quick feedback.
That makes sense.

I'll close this PR.

Thank you.

daniel22-jung-ivis:
Although I closed the PR, I realized that I missed mentioning the `.proto` file.
I am including the `.proto` definition here for the record:


```protobuf
enum DeviceStatus2 {
    option (nanopb_enumopt).enum_intsize = IS_8; 
    DeviceStatus2_OFFLINE = 0; 
    DeviceStatus2_ONLINE  = 1; 
    DeviceStatus2_ERROR   = 2;
}
```


If you decide this is worth addressing, please leave a comment, and I will reopen the PR.

